PR #6, opened 2018-12-24
Add version and git sha in Constants.java

thescouser89:
`Constants.java` is now autogenerated from a templates file where the
variables are substituted with the real value. To make it work
seemlessly to capture the real SHA used and version, the generated
`Constants.java` is 'git ignored'; only the changes in the template file
is tracked


thescouser89:
The version and sha in `Constants.java` can be used to set the version in the CLI help. I think this will be very useful when helping CLI users to know the version and SHA used to help debug.

PR #8, opened 2018-12-24
Use PNC's codeformat style

thescouser89:
This is based on the eclipse codestyle from Dependency Analysis

### DIFF
--- a/src/main/java/org/jboss/pnc/bacon/App.java
+++ b/src/main/java/org/jboss/pnc/bacon/App.java
@@ -31,11 +31,10 @@ import java.util.List;
  * <br>
  * Date: 12/13/18
  */
-@CommandLine.Command(name = "java -jar bacon.jar",
-        mixinStandardHelpOptions = true,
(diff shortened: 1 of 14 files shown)


thescouser89:
I wouldn't blame you if you reject that PR based on the formatted code. Some of the formatting for the annotatiions are awful. I wonder if we can fix that in the future.

src/main/java/org/jboss/pnc/bacon/App.java
```diff
@@ -31,11 +31,10 @@
  * <br>
  * Date: 12/13/18
  */
-@CommandLine.Command(name = "java -jar bacon.jar",
-        mixinStandardHelpOptions = true,
-        subcommands = {Pig.class, Pnc.class, Da.class}
-)
+@CommandLine.Command(name = "java -jar bacon.jar", mixinStandardHelpOptions = true, subcommands = {
```
michalszynkiewicz: what happened here? previous formatting was better

PR #14, opened 2019-02-07
Use keycloak config

thescouser89:


src/main/java/org/jboss/pnc/bacon/config/KeycloakConfig.java
```diff
@@ -0,0 +1,23 @@
...
+public class KeycloakConfig {
+
+    // Url of Keycloak server
+    private String url;
+
+    // Realm of keycloak
+    private String realm;
+
+    // Client to use to authenticate. Can be a regular client or a client service account
+    private String clientId;
```
michalszynkiewicz: do you think we could provide some defaults from clientId and realm?

thescouser89: Hmm maybe but nothing that would make sense in the opensource code. We just picked some values in our Keycloak instance.

PR #16, opened 2019-02-19
[NCL-4338] Auth using direct flow for Keycloak

thescouser89:
This commit adds support for authentication to Keycloak using direct
flow. We are also working on using official Keycloak libraries to get
authentication to work via authorization code flow (which is more secure)

It also adds an interface for authentication which can be used for both
direct flow, authorization code flow, and for any other kind of flow in
the future.

## TODO
- Save the access token and the refresh token in a cache file somewhere
  to re-use if still valid.

cc: @matejonnet @michalszynkiewicz @jbartece @geoand

### DIFF
--- a/auth/src/main/java/org/jboss/pnc/bacon/auth/DirectKeycloakClientImpl.java
+++ b/auth/src/main/java/org/jboss/pnc/bacon/auth/DirectKeycloakClientImpl.java
@@ -1,0 +1,141 @@
+package org.jboss.pnc.bacon.auth;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.mashape.unirest.http.HttpResponse;
+import com.mashape.unirest.http.ObjectMapper;
+import com.mashape.unirest.http.Unirest;
+import lombok.extern.slf4j.Slf4j;
+import org.jboss.pnc.bacon.auth.model.Credential;
+import org.jboss.pnc.bacon.auth.model.KeycloakResponse;
+import org.jboss.pnc.bacon.auth.spi.KeycloakClient;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Date;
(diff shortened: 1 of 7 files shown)


auth/src/main/java/org/jboss/pnc/bacon/auth/DirectKeycloakClientImpl.java
```diff
@@ -0,0 +1,141 @@
...
+                try {
+                    return jacksonObjectMapper.writeValueAsString(value);
+                } catch (JsonProcessingException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        });
+    }
+
+    public DirectKeycloakClientImpl() {
```
michalszynkiewicz: Could `setupUnirest()` be done in static block? Just in case someone creates many instances of this class

auth/src/main/java/org/jboss/pnc/bacon/auth/DirectKeycloakClientImpl.java
```diff
@@ -0,0 +1,141 @@
...
+            log.debug("Getting token via username/password");
+
+            HttpResponse<KeycloakResponse> postResponse = Unirest.post(keycloakEndpoint)
+                    .field("grant_type", "password")
+                    .field("client_id", client)
+                    .field("username", username)
+                    .field("password", password)
+                    .asObject(KeycloakResponse.class);
+
+            KeycloakResponse response = postResponse.getBody();
```
michalszynkiewicz: if it fails, will it show some pretty exception containing the status code of response, etc?

thescouser89: I'm not sure actually. WIll check

PR #26, opened 2019-04-04
Make our logback settings *the* logback setting

thescouser89:
The logback.xml in koji-build-finder is 'erasing' our cli's logback.xml
in the final jar. To workaround this, this commit renames our
logback.xml to logback-test.xml, since this has higher priority compared
to logback.xml

https://logback.qos.ch/manual/configuration.html


thescouser89:
If you guys know of another way to make our app logback.xml win over the koji-build-finder logback.xml, let me know!

PR #32, opened 2019-07-01
pnc cli -> pnc 2.0 rest

michalszynkiewicz:


michalszynkiewicz:
@thescouser89 yeah, sorry about the size.
Thank you for the review!

thescouser89:
No worries about the size! In this case it has to be the way it has to be :)

PR #37, opened 2019-08-08
[NCL-4338] Add Keycloak authentication

thescouser89:
The Keycloak authentication automatically kicks in when the config file
has the keycloak section in it.

The authentication is currently being done via the direct flow.

A helper method in `PncClientHelper#authRequired` is provided to *stop*
the application if it did not authenticate to Keycloak.

The format of the keycloak config is specified in the config.yaml


thescouser89:
@michalszynkiewicz Could you give it a quick review please? :)

PR #47, opened 2019-09-30
Add a retry handler to handle rate limiting routes

vibe13:


vibe13:
After we introduced the rate limiting request to Indy, CI jobs started encountering org.apache.http.NoHttpResponseException at org.jboss.prod.generator.utils.FileDownloadUtils.downloadWithClient(FileDownloadUtils.java:101).
This patch tries to mitigate such errors with some retries (up to 5 times). cc @thescouser89 @michalszynkiewicz wdyt? Thanks!

thescouser89:
That makes a lot of sense @vibe13 ! Change looks good!

PR #62, opened 2019-10-11
[NCL-5206] Don't store krb password in main config.yaml

thescouser89:
This commit removes the requirement of specifying the Kerberos password
in the config.yaml. Instead, it asks for the password.

In addition, caching of the access / refresh tokens is added to speed up
authentication on subsequent requests.

Keycloak tells us how much time the refresh and access tokens are valid.
If the access token will expire in less than 20 hours, we'll use the
refresh token to request a new access/refresh tokens and save it in our
cached location.

If the refresh token is already expired, we'll just go back to ask the
user for their password, as is the case for the PNC Web UI.

The cached tokens are invalidated if the config.yaml has an altered
keycloak url, username, realm. This is because those tokens
can't be used with the different settings.

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -13,6 +13,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.jboss.pnc.bacon</groupId>
+            <artifactId>common</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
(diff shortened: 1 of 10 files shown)


thescouser89:
Examples:
- Entering password:
```
➜  cli git:(ncl-5206) ✗ java -jar cli/target/bacon.jar pnc product list
[DEBUG] - clientSecret is not specified in the config file! Assuming this is a regular user
[DEBUG] - Authenticating to keycloak
[DEBUG] - Getting token via username/password
Enter your password: 
[DEBUG] - Writing credential to cache file
<output>
```

- Entering wrong password
```
➜  cli git:(ncl-5206) ✗ java -jar cli/target/bacon.jar pnc product list
[DEBUG] - clientSecret is not specified in the config file! Assuming this is a regular user
[DEBUG] - Authenticating to keycloak
[DEBUG] - Getting token via username/password
Enter your password: 
[DEBUG] - Writing credential to cache file
[ERROR] - Credentials don't seem to be valid
```

- Using cached file after initial successful authentication:
```
➜  cli git:(ncl-5206) ✗ java -jar cli/target/bacon.jar pnc product list
```
[DEBUG] - clientSecret is not specified in the config file! Assuming this is a regular user
[DEBUG] - Authenticating to keycloak
[DEBUG] - Using cached credential details
<output>
```

PR #85, opened 2019-11-19
[NCL-4287] Add example section if user overrides exampleText method

thescouser89:
ncl-4287
ncl-4287-2


thescouser89:
retest this please

PR #87, opened 2019-11-20
Update for changed DTOs

janinko:


thescouser89:
Thanks for the PR Sensei @janinko ! :)

PR #92, opened 2019-12-02
Adapt to buildLog now being a stream

thescouser89:
In PNC 2.0, the build log has been changed to use streams so as not to
consume too much memory. As such, we need to adjust our codebase


thescouser89:
Change on PNC side is here: https://github.com/project-ncl/pnc/commit/e5ed237f37c229fc209bf5c2a2b6b0151b92ef27

thescouser89:
Fixed in #94

PR #111, opened 2020-02-04
[NCL-5444] CLI cache file for token should have permission 600

jomrazek:
Commit changes saved-used.json cache file permissions to "-rw-------"


auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -31,18 +34,19 @@
     }
 
     public static void writeCredentialToCacheFile(String keycloakUrl, String realm, String username, Credential credential) {
-        log.debug("Writing credential to cache file");
 
         createConfigFolderIfAbsent();
```
jomrazek: I was expecting to see log about creating config folder before log about writing to config file, that's reason for moving log output

PR #113, opened 2020-02-06
[NCL-5451] Allow user to specify format of output: YAML or JSON

jomrazek:
Commit changes default output for commands with relevant info to yaml, using `-o/--jsonOutput `output could be changed to json format.

### DIFF
--- a/common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractGetSpecificCommand.java
+++ b/common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractGetSpecificCommand.java
@@ -22,6 +22,8 @@ import org.aesh.command.CommandException;
 import org.aesh.command.CommandResult;
 import org.aesh.command.invocation.CommandInvocation;
 import org.aesh.command.option.Argument;
+import org.aesh.command.option.Option;
+import org.jboss.pnc.bacon.common.ObjectHelper;
 import org.jboss.pnc.client.ClientException;
 
 /**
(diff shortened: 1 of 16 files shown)


common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractGetSpecificCommand.java
```diff
@@ -36,10 +38,15 @@
...
 
+    @Option(shortName = 'o', overrideRequired = false, hasValue = false, description = "use json for output (default to yaml)")
+    private boolean jsonOutput = false;
+
     @Override
     public CommandResult execute(CommandInvocation commandInvocation) throws CommandException, InterruptedException {
 
-        return super.executeHelper(commandInvocation, () -> System.out.println(getSpecific(id)));
+        return super.executeHelper(commandInvocation, () -> {
+            System.out.println(ObjectHelper.getOutputMapper(jsonOutput).writeValueAsString(getSpecific(id)));
```
thescouser89: I think we could shorten the method call to:

```
ObjectHelper.print(<is json output?>, object);
```

It's shorter and it hides the implementation details of the formatter engine we are using (Jackson).

thescouser89: Let me know what you think! :)

jomrazek: :+1:

PR #117, opened 2020-02-10
[NCL-5465] Edit error messages to more elegant and user friendly format

jomrazek:
Mainly made sure errors are logged and displayed without stacktrace for pnc commands, I also went through pig to be sure nothing is just printed out, didn't remove logging of stacktrace  (don't know if it can be helpful for user in pig, open to suggestions)

### DIFF
--- a/config/src/main/java/org/jboss/pnc/bacon/config/Config.java
+++ b/config/src/main/java/org/jboss/pnc/bacon/config/Config.java
@@ -65,10 +65,8 @@ public class Config {
     public static void initialize() throws IOException {
         File configFile = new File(configLocation);
         if (!configFile.exists()) {
-            log.error(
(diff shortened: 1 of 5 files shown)


config/src/main/java/org/jboss/pnc/bacon/config/Config.java
```diff
@@ -65,10 +65,8 @@ public static void configure(String configLocation) throws IOException {
     public static void initialize() throws IOException {
         File configFile = new File(configLocation);
         if (!configFile.exists()) {
-            log.error(
-                    "Config file: '{}' does not exist. Please create a config file and either name it 'config.yaml' and put it in the working directory or specify it with -Dconfig",
-                    configLocation);
-            throw new IOException("Config file " + configLocation + " does not exist!");
```
jomrazek: We were logging missing of config file twice, once here and once when catching exception

jomrazek:
retest this please

thescouser89:
Thanks for the changes! It does make a lot of sense!

PR #120, opened 2020-02-17
Config file for PNC can be set through flag or ENV

dominikbrazdil:


common/src/main/java/org/jboss/pnc/bacon/common/Path.java
```diff
@@ -0,0 +1,8 @@
+package org.jboss.pnc.bacon.common;
+
+public class Path {
```
jomrazek: Conflicting with java.nio.file.Path, for other use I would prefer name to be changed as we discussed, for this change I checked code and there already is Config class containing config location. You can improve our code and unify usages and setup of config file.

common/src/main/java/org/jboss/pnc/bacon/common/Path.java
```diff
@@ -0,0 +1,8 @@
...
+
+    public static String CONFIG_FOLDER = System.getProperty("user.home") + "/.config/pnc-bacon";
```
jomrazek: when it's not final anymore shouldn't it be camel case?

common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractCommand.java
```diff
@@ -55,6 +56,9 @@
     @Option(shortName = 'v', overrideRequired = true, hasValue = false, description = "Verbose output")
     private boolean verbose = false;
 
+    @Option(shortName = 'p', description = "Path to configuration file")
```
jomrazek: Test it before creating PR :P

common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractCommand.java
```diff
@@ -96,6 +100,22 @@ private void setVerbosityIfPresent() {
         }
     }
 
+
+    /**
+     * Set the path to config file if the configPath flag or environment variable is set
+     *
+     */
+    private void setConfigurationFileLocationIfPresent() {
```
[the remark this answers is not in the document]
dominikbrazdil: It was supposed to be in executePrivate() method and it was there when i was testing it. I must have accidentally removed it somehow before commit.

common/src/main/java/org/jboss/pnc/bacon/common/Path.java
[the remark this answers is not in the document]
dominikbrazdil: I was little bit confused about which config is for what and I was trying to setup the Config.configLocation first, but since I need the flag parameter to be set in AbstractCommand I can't access the static part of the configuration (e.g. Config or CacheFile).

auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -23,7 +23,7 @@
 @Slf4j
 public class CacheFile {
 
-    public static final String CACHE_FILE = Constant.CONFIG_FOLDER + "/" + "saved-user.json";
+    public static final String CACHE_FILE = Config.getConfigLocation() + "/" + "saved-user.json";
```
janinko: This won't work if the config location is changed after program start - it will still point to the original location.

auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -75,9 +75,9 @@ public static void writeCredentialToCacheFile(String keycloakUrl, String realm,
     }
 
     private static void createConfigFolderIfAbsent() {
-        if (!fileExists(Constant.CONFIG_FOLDER)) {
+        if (!fileExists(Constant.DEFAULT_CONFIG_FOLDER)) {
```
janinko: It should create the folder where the saved-user.json will be actually stored, not the default location.

auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -100,7 +100,7 @@ private static String generateUsernameMd5(String keycloakUrl, String realm, Stri
 
     private static boolean fileExists(String pathString) {
 
-        Path path = Paths.get(pathString);
+        java.nio.file.Path path = Paths.get(pathString);
```
janinko: I think nothing now prevents the `java.nio.file.Path` to be imported so you can use only the name `Path` without the package here.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -103,9 +104,8 @@ public static void main(String[] args) throws Exception {
     }
 
     private static void initializeConfig() {
-        String configLocation = System.getProperty("config", "config.yaml");
         try {
-            Config.configure(configLocation);
+            Config.configure(Constant.DEFAULT_CONFIG_FOLDER, Constant.CONFIG_FILE_NAME);
```
janinko: I would maybe move this to the `setConfigurationFileLocationIfPresent` method so all three ways are at one place (and then renamed the method just `setConfigurationFileLocation`)

common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractCommand.java
```diff
@@ -96,6 +103,31 @@ private void setVerbosityIfPresent() {
...
+            setConfigLocation(System.getenv(Constant.CONFIG_ENV), "environment variable");
+        }
+    }
+
+    private void setConfigLocation(String configLocation, String source) {
+        try {
+            Config.configure(configLocation, Constant.CONFIG_FILE_NAME);
+            log.debug("Config file set from " + source + " to " + Config.getConfigLocation());
+        } catch (IOException e) {
+            System.err.println("Configuration file not found. "
```
janinko: Use `log.warning` or `log.error` instead of `System.err`

config/src/main/java/org/jboss/pnc/bacon/config/Config.java
```diff
@@ -58,22 +59,29 @@ public static Config instance() throws RuntimeException {
         return instance;
     }
 
-    public static void configure(String configLocation) throws IOException {
+    public static void configure(String configLocation, String configFileName) throws IOException {
```
janinko: This method doesn't throw `IOException` ATM.

config/src/main/java/org/jboss/pnc/bacon/config/Config.java
```diff
@@ -58,22 +59,29 @@ public static Config instance() throws RuntimeException {
...
+        Config.configFileName = configFileName;
     }
 
     public static void initialize() throws IOException {
-        File configFile = new File(configLocation);
+        File configFile = new File(configLocation + "/" + configFileName);
         if (!configFile.exists()) {
             throw new IOException("Config file " + configLocation
-                    + " does not exist! Please create a config file and either name it 'config.yaml' and put it in the working directory or specify it with -Dconfig");
+                    + " does not exist! Please create a config file and either name it 'config.yaml' and "
```
janinko: you now have `configFileName` variable, you shoul use it in the message

config/src/main/java/utils/Validator.java
```diff
@@ -0,0 +1,24 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Validator {
+
+    public static void checkIfNull(Object object, String reason) {
+        if (object == null) {
+            log.error(reason);
+            throw new ConfigMissingException();
```
janinko: It would be good if the exception also takes the reason.

config/src/main/java/utils/Validator.java
```diff
@@ -0,0 +1,24 @@
+package utils;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Validator {
```
janinko: Maybe move the content of this class to the Validate interface?

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -26,6 +26,7 @@
 import org.aesh.command.parser.RequiredOptionException;
 import org.aesh.command.registry.CommandRegistry;
 import org.jboss.bacon.da.Da;
+import org.jboss.pnc.bacon.common.Constant;
```
jomrazek: is it needed? probably just leftover from previous solution

PR #127, opened 2020-02-20
[NCL-5509] Fix Authentication broken in multicall commands

thescouser89:
This is done by getting rid of the duplicate 'getClient' and
'getClientAuthenticated' methods in the various Cli classes and
replacing it with the `ClientCreator` class which contains the logic.

It uses reflection to create a new instance of the Client and properly
caches the un-authenticated, and authenticated clients.

### DIFF
--- a/pnc/src/main/java/org/jboss/pnc/bacon/pnc/common/ClientCreator.java
+++ b/pnc/src/main/java/org/jboss/pnc/bacon/pnc/common/ClientCreator.java
@@ -1,0 +1,63 @@
+package org.jboss.pnc.bacon.pnc.common;
+
+import lombok.extern.slf4j.Slf4j;
+import org.jboss.pnc.bacon.pnc.client.PncClientHelper;
+import org.jboss.pnc.client.ClientBase;
+import org.jboss.pnc.client.Configuration;
+
+import java.util.function.Function;
+
+/**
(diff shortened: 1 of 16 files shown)


thescouser89:
retest this please

pnc/src/main/java/org/jboss/pnc/bacon/pnc/common/ClientCreator.java
```diff
@@ -0,0 +1,72 @@
...
+    private Class<T> tClass;
+
+    private T client;
+    private T clientAuthenticated;
+
+    /**
+     * While Java Generics is nice, there's no way of getting the class object of a generic type. As such, we need to find a way
+     * to capture the Generic class, hence the constructor to capture it. It's sad I know
+     */
+    public ClientCreator(Class<T> tClass) {
```
janinko: If you don't want to use reflection, you can accept `Supplier<T>` for the constructor method and then create ClientCreator as:
```
new ClientCreator<>(ProjectClient::new);
```

Not sure if that is better, but it's an option.

thescouser89: I'll try it. Getting rid of reflection is better in case we want to compile it with Graal. Thanks for the suggestion buddy! :)

PR #132, opened 2020-02-26
Fix ClientCreator

janinko:


thescouser89:
That's an embarrassing bug. Thanks for fixing it Honza!

PR #166, opened 2020-03-23
Fix eg config.yaml alignment for keycloak and check

thescouser89:
Add check to make sure that the keycloak section in the configuration is
filled if we require authentication.

e.g when the Keycloak section is not specified in the configuration file
```
➜  bacon git:(fix-config) ✗ java -jar cli/target/bacon.jar pnc build start 100 -v -p ~/
[DEBUG] - Log level set to DEBUG
[DEBUG] - Config file set from flag to /home/dcheung//config.yaml
[ERROR] - Keycloak section is needed in the configuration file!
```

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -21,39 +21,37 @@ profile:
   pnc:
       url: ""
       bifrostBaseurl: ""
-
 # ******************************************************************************
 # Authentication information
 #
 # Uncomment this section if you want to create / update / delete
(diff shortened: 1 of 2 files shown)


thescouser89:
CC: @michalovjan this PR should fix the issues we had earlier. Thank you for reporting them so that we can improve bacon! :)

PR #186, opened 2020-04-30
[NCL-5445] Add bash completion support for CLI 2.x

jomrazek:
Native bash completion is not straight forward as it might have seem so I decided to use aesh console to simplify and add reliability to this issue.

Bacon usage can now be split into 2 cases
-bacon console (for comfort manual usage with completion of available commands)
when running bacon without arguments `java -jar bacon.jar` it starts bacon console where you can specify commands directly without running java -jar bacon.jar for every command.
bacon-console

-script usage (bacon how it functioned until now)
adding arguments to command will cause to not start bacon console and allow bacon to be still used as it was until now
bacon

possible improvements: loading config and profile for whole console session... ?

### DIFF
--- a/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
+++ b/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
@@ -21,14 +21,27 @@ import java.util.Arrays;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.aesh.command.AeshCommandRuntimeBuilder;
+import org.aesh.command.CommandResult;
 import org.aesh.command.CommandRuntime;
 import org.aesh.command.GroupCommandDefinition;
+import org.aesh.command.activator.CommandActivator;
+import org.aesh.command.activator.OptionActivator;
+import org.aesh.command.builder.CommandBuilder;
+import org.aesh.command.completer.CompleterInvocation;
+import org.aesh.command.converter.ConverterInvocation;
 import org.aesh.command.impl.registry.AeshCommandRegistryBuilder;
+import org.aesh.command.invocation.CommandInvocation;
 import org.aesh.command.parser.CommandLineParserException;
 import org.aesh.command.parser.OptionParserException;
 import org.aesh.command.parser.RequiredOptionException;
 import org.aesh.command.registry.CommandRegistry;
-import org.apache.commons.lang.StringEscapeUtils;
+import org.aesh.command.settings.SettingsBuilder;
+import org.aesh.command.validator.ValidatorInvocation;
+import org.aesh.readline.Prompt;
+import org.aesh.readline.ReadlineConsole;
+import org.aesh.readline.terminal.formatting.Color;
+import org.aesh.readline.terminal.formatting.TerminalColor;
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -41,41 +53,69 @@
  */
 
 @Slf4j
-@GroupCommandDefinition(name = "bacon", description = "Bacon CLI", groupCommands = { Pnc.class, Da.class, Pig.class })
```
jomrazek, line 44: anybody know reasoning for this to be here and then appending it on line 77?

thescouser89:
PR looks ok! I'm conflicted on whether to enable the console mode while running 
```
java -jar bacon.jar
```

vs

```
java -jar bacon.jar console
```

But I think this change makes sense. When you type bacon.jar without any flags, you'll just launch the app in console mode. I'm willing to give it a try and see if users get confused by it.

Unfortunately the removal of `@GroupCommandDefinition` also removed the option to parse '-h' or '-V':

```
➜  bacon git:(jomrazek-NCL-5445) java -jar cli/target/bacon.jar --help
Exception in thread "main" org.aesh.command.CommandNotFoundException: Command: named --help was not found.
	at org.aesh.command.impl.registry.MutableCommandRegistryImpl.getCommandByAlias(MutableCommandRegistryImpl.java:214)
	at org.aesh.command.impl.AeshCommandResolver.getCommand(AeshCommandResolver.java:92)
	at org.aesh.command.impl.AeshCommandResolver.resolveCommand(AeshCommandResolver.java:60)
	at org.aesh.command.impl.AeshCommandRuntime.findCommandContainer(AeshCommandRuntime.java:229)
	at org.aesh.command.impl.Executions.buildExecution(Executions.java:325)
	at org.aesh.command.impl.AeshCommandRuntime.buildExecutor(AeshCommandRuntime.java:214)
	at org.aesh.command.impl.AeshCommandRuntime.executeCommand(AeshCommandRuntime.java:147)
	at org.jboss.pnc.bacon.cli.App.run(App.java:93)
	at org.jboss.pnc.bacon.cli.App.main(App.java:124)
```

I guess the annotation does add the '--help' flag automagically. Could the console be activated / configured without having to remove the `@GroupCommandDefinition`?

The other tiny suggestion is to print a banner when entering the 'console' mode to let the user know that they are now running bacon in 'console' mode and pnc commands will be tab-completed. I think it might confuse users to see a shell very similar to their own shell and think they are in Bash.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
jomrazek: got it, thank you! (before)

jomrazek:
@thescouser89 thanks for feedback, fixed back that group command and splitted command registry so it's not showing up in console, unfortunately it's called readline console for a reason there seems to be no easy way to add banner or print something at the start when I looked at it but I like your idea and it makes sense so I will try to figure out.

thescouser89:
Thanks for fixing it! LGTM! :)

PR #198, opened 2020-05-21
Updates to README.md for PiG and PNC

thescouser89:


README.md
```diff
@@ -44,3 +69,54 @@ keycloak:
...
 ```
+
+---
+
+# PiG
+PiG allows users to setup PNC with the product, build configurations, and much more via a YAML file. This makes it much easier to configure PNC for new releases as well as making the configuration portable across different PNC servers.
+
+PiG is run in 2 phases:
+
+- configure
```
michalszynkiewicz, line 80: there are more.
Generally there are two ways to running this stuff:
- all in one, `bacon pig run -c ...`
- single step `bacon pig <step> -c...`

The steps are `configure`, `build`, `repo`, `licenses`, `javadocs`, ...

thescouser89: @michalszynkiewicz `bacon` doesn't have the `run` part. Are you planning to add it at some point?

Thanks for the input! I'll update the doc

michalszynkiewicz: What do you mean it doesn't have the run part? 
The command is defined: https://github.com/project-ncl/bacon/blob/master/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java#L95

thescouser89: Ah! It's just not added in the group commands, so that's why it doesn't show up. Will create PR to add it and add it in the documentation.

thescouser89: Added and merged! `run` is now available :)

PR #201, opened 2020-05-25
Fix pig errors

thescouser89:
Test yaml used:
```yaml
product:
  name: ProductB # full product name
  abbreviation: testb # official product name abbreviation - used to create the brew/koji tag name
  stage: Alpha1 # product phase, if we're targetting GA, should be GA, etc
  issueTrackerUrl: http://issues.jboss.org/browse/TEST
version: 1.2.0
milestone: ER1
group: ProductA 1.2 all # name of the PNC build group
```


thescouser89:
Thanks for the review @michalszynkiewicz ! :)

PR #260, opened 2020-07-21
Bug fix 3

thescouser89:


thescouser89:
Hi @michalszynkiewicz ! Could you please review the last commit please?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/GitRepoInspector.java
```diff
@@ -122,7 +122,7 @@ private static String headRevision(Git git, String branch) throws GitAPIExceptio
     }
 
     private static String getLatestBuiltRevision(String configId) {
-        PncBuild latestBuild = buildInfoCollector.getLatestBuild(configId);
+        PncBuild latestBuild = buildInfoCollector.getLatestBuild(configId, BuildInfoCollector.BuildSearchType.ALL);
         return latestBuild.getBuildConfigRevision().getScmRevision();
```
michalszynkiewicz, line 126: The biild search type here should correspond to the type of build we're doing

thescouser89: Ack! Will fix this

thescouser89: Re-reading the feature, it only checks if the latest build has the commit id of the latest in the branch. As such, I believe it doesn't matter if the latest successful build is a temporary build or a permanent build since the pre-build sync will happen. Let me know if you disagree! :)

michalszynkiewicz: Why?

The problem we're solving here is that PNC doesn't consider adding commits to a branch a change that should cause triggering a rebuild.
For this reason, PiG adds a parameter for a build if it finds out that a branch that is used to build the project has more commits since the previous build.
If we're doing a permanent build, we should take into account the previous permanent build in the check. 
Am I missing something?

thescouser89: Ah gotcha gotcha! Thanks for clarifying! :)

PR #266, opened 2020-07-27
Check if branch modified for temp or permanent build

thescouser89:


pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
```diff
@@ -254,9 +254,17 @@ public String doExecute() {
                 description = SKIP_BRANCH_CHECK_DESC)
         private boolean skipBranchCheck;
 
+        @Option(
```
michalszynkiewicz, line 257: `tempBuild` is already defined in `PigCommand`, I don't think it's needed here

michalszynkiewicz: Feel free to merge when this is fixed, the rest looks good.
Thank you!

thescouser89: Ah thanks for letting me know! I didn't even check! :)

PR #268, opened 2020-07-29
[NCL-5882] Allow user to specify systemImgaeId instead of environmentId

thescouser89:


thescouser89:
Documentation: https://github.com/project-ncl/bacon/wiki/PiG-features#portability-of-build-configyaml-between-different-pnc-environments

PR #273, opened 2020-08-03
check if pig configured

michalszynkiewicz:


michalszynkiewicz:
@thescouser89 I'm merging it but please take a look if you have a while

PR #284, opened 2020-08-07
Force formatter-maven-plugin lineEnding to LF for consistency

dwalluck:


dwalluck:
This fixes an issue with formatter changing all line endings to system line endings which can affect Windows which uses CRLF (this assumes that all existing line ending are Unix/LF, which they are).

PR #323, opened 2020-08-20
Fix Koji Build Finder configuration file loading and pool shutdown

dwalluck:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
```diff
@@ -43,8 +44,7 @@ private BuildFinderUtils() {
 
     public static BuildConfig getKojiBuildFinderConfigFromFile(File file) {
         try {
-            final String json = org.apache.commons.io.FileUtils.readFileToString(file, StandardCharsets.UTF_8);
-            return getKojiBuildFinderConfigFromJson(json);
+            return BuildConfig.load(file);
```
dwalluck, line 47: Use built-in load method which takes a `File`.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
```diff
@@ -135,7 +135,7 @@ public static BuildConfig getKojiBuildFinderConfig() {
             log.error("Koji client error: {}", e.getMessage(), e);
             return Collections.emptyList();
         } finally {
-            pool.shutdown();
+            Utils.shutdownAndAwaitTermination(pool);
```
dwalluck, line 138: This does more than just `shutdown` in accordance with the documentation at https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html.

rnc (approved):

PR #330, opened 2020-08-21
Fix path parsing inside tar archives

dwalluck:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/sources/SourcesGenerator.java
```diff
@@ -120,7 +120,9 @@ private void downloadSourcesFromBuilds(Map<String, PncBuild> builds, File workDi
 
     private boolean isNotANestedFile(String name) {
         // either "some-directory" or "some-directory/"
-        return !name.contains(SEPARATOR) || name.indexOf(SEPARATOR) + SEPARATOR.length() == name.length();
+        // paths inside tar files only ever contain forward slashes
+        int i = name.indexOf("/");
```
dwalluck, line 124: Avoid scanning for "/" twice.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/sources/SourcesGenerator.java
```diff
@@ -106,8 +105,9 @@ private void downloadSourcesFromBuilds(Map<String, PncBuild> builds, File workDi
 
             if (topLevelDirectories.size() != 1) {
                 throw new RuntimeException(
-                        "Invalid number of top level directories untared for build " + build + ", "
-                                + "the untared archive: " + targetPath.getAbsolutePath());
+                        "Found more than one top-level directory (" + topLevelDirectories.size()
+                                + ") untared for build " + build + ", the untared archive: "
+                                + targetPath.getAbsolutePath() + ", the top level directories:" + topLevelDirectories);
```
dwalluck, line 110: Add more details to error message.

rnc:
Should we have a test here?

dwalluck:
> (For reference , I have plenty of unit tests in pme that test private methods using reflection)

It seems that Mockito recently added the ability to mock static classes, which is nice because I just needed to use it. Unfortunately, it still can't mock private classes. But, I really think this method might better be made `public` and moved to `pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileUtils.java`.

PR #336, opened 2020-08-21
[#338] Fix issues getting version

dwalluck:
* Close input stream
* Check for null pointers
* Don't create map to store instance
* Handle multiple jars on classpath


rnc:
IMHO when this code already exists in dependent projects (e.g. PME) we should reuse that code and avoid duplication.

dwalluck:
> Version #355

I guess you can only link to an issue and not another PR? Huh.

thescouser89:
retest this please

michalszynkiewicz:
@rnc says it's superseded by the merged #355 355

PR #346, opened 2020-08-24
Fix swapped argument order in assert statements in ObjectHelperTest

dwalluck:
Fix mix of charsets UTF-8 and default in ObjectHelperTest


dwalluck:
The `ObjectMapper` outputs using UTF-8 charset and seems to use `\n` line ending regardless of platform, but `PrintStream` uses the platform's default encoding and line ending by default and can lead to test failures in `printJson()` which can get a mix of two different line endings and encodings in the same string,

Issue #326, opened 2020-08-20
Unable to run PiG with --tempBuild specified


goldmann:
```
$ bacon pig run --tempBuild /workDir/workspace/pnc-2.0/test-java/cpaas-test-java/cpaas-test-pig-project/config/cpaas/development
[...]

PR #348, opened 2020-08-24
[NCL-5943] fix NoSuccessfulBuildException on unchanged BuildConfig when a temporary build is not rebuilt after a regular build

michalovjan:
Hi guys. This should fix NCL-5943(#326) but as I am not familiar with PIG I fear this specific solution($1) might have side-effects which I cannot predict. 

Other possible solution($2) is to handle this case just before throwing an exception here:
https://github.com/project-ncl/bacon/blob/869832e483c9fd9564b5474054394c92e38a71c8/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/BuildInfoCollector.java#L89

Should I go with $1, try to improve $1 or go with $2


thescouser89:
Based on the disccusions in the JIRA, approved.

PR #355, opened 2020-08-24
Version

rnc:


rnc:
Supercedes #336 

@thescouser89 The change to minimal in hindsight isn't correct if PME is actually being executed. This reverts that, using the reverted code to simplify the version handling.

michalszynkiewicz (approved):

PR #363, opened 2020-08-25
Remove withJansi from logback-test.xml

dwalluck:
The default value is false. There is no need to change it since it's not required on modern platforms and it breaks the tests. And it won't work unless you also have jansi on the classpath.


thescouser89:
LGTM!

PR #364, opened 2020-08-25
Drop PME's jhttpc

michalszynkiewicz:


michalszynkiewicz:
CC @rnc maybe you could also update the jhttpc version used by PME?

rnc:
@michalszynkiewicz We don't directly depend on it ; it looks like it comes from Galley - will investigate!

rnc:
@michalszynkiewicz Which version of jhttpc do you specifically need?

rnc:
@dwalluck Do you know what the minimum version of jhttpc KBF needs is? It looks like Galley has 1.9, and while KBF has been updated to the latest jhttpc I don't know if it _requires_ it?

dwalluck:
> @dwalluck Do you know what the minimum version of jhttpc KBF needs is? It looks like Galley has 1.9, and while KBF has been updated to the latest jhttpc I don't know if it _requires_ it?

Technically, it should only need whatever version is being used by the latest kojiji which appears to be 1.6. So any version 1.6 or later apparently works.

michalszynkiewicz:
KBF uses 1.11.
The method that I think was missing from the version shaded in PME was `withConnectionPoolTimeoutSeconds`

PR #365, opened 2020-08-26
Add ability to find checksums and builds separately

dwalluck:
Requested by @michalszynkiewicz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -308,6 +308,12 @@
                 <groupId>org.commonjava.maven.ext</groupId>
                 <artifactId>pom-manipulation-common</artifactId>
                 <version>4.0</version>
+                <exclusions>
+                  <exclusion>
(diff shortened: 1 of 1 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/PigContext.java
```diff
@@ -83,6 +84,8 @@
 
     private String fullVersion; // version like 1.3.2.DR7
 
+    private Map<String, Collection<String>> checksums;
```
michalszynkiewicz, line 87: (not for this commit) it would be good to have some more meaningful name. Sth like repositoryContentsChecksums

michalszynkiewicz:
I screwed up rebasing and created a separate PR with the changes (rebased): https://github.com/project-ncl/bacon/pull/370

PR #366, opened 2020-08-26
Close PNC clients in PigFacade

dwalluck:
There are still many places in the code where clients still aren't being closed. One place where it's difficult to close them is when they are initialized in the constructor of a class.


dwalluck:
retest this please

PR #367, opened 2020-08-26
Fix checking for existing SCM Repository in pig

janinko:
Don't use buildConfig::matchesRepository in the check as PNC's match endpoint
should be always returning single matching SCM Repository if it exists.

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncEntitiesImporter.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncEntitiesImporter.java
@@ -344,7 +344,16 @@ public class PncEntitiesImporter {
         }
 
         try {
-            return toStream(repoClient.getAll(matchUrl, null)).filter(buildConfig::matchesRepository).findAny();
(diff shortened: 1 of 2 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncEntitiesImporter.java
```diff
@@ -344,7 +344,14 @@ private SCMRepository getOrGenerateRepository(BuildConfig buildConfig) {
...
 
         try {
-            return toStream(repoClient.getAll(matchUrl, null)).filter(buildConfig::matchesRepository).findAny();
+            List<SCMRepository> foundRepository = toStream(repoClient.getAll(matchUrl, null)).collect(Collectors.toList());
+            if (foundRepository.isEmpty()) {
+                return Optional.empty();
+            } else if (foundRepository.size() == 1) {
+                return Optional.of(foundRepository.get(0));
+            } else {
+                throw new RuntimeException("There exists more then one SCM Repository for url: " + matchUrl + " " + foundRepository);
```
janinko, line 353: This should never happen (but I didn't want to write this message there to not embarass myself :D)

michalszynkiewicz (approved):

PR #372, opened 2020-08-26
Update to build-finder 2.0.0-SNAPSHOT.

matejonnet:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
```diff
@@ -172,10 +171,10 @@ public static BuildConfig getKojiBuildFinderConfig() {
      */
     public static List<KojiBuild> findBuilds(File file, boolean includeNotFound) {
         BuildConfig config = getKojiBuildFinderConfig();
-        List<File> inputs = Collections.singletonList(file);
-        ExecutorService pool = Executors.newFixedThreadPool(2);
+        List<String> inputs = Collections.singletonList(file.getPath());
+        ExecutorService pool = Executors.newFixedThreadPool(config.getChecksumTypes().size());
```
dwalluck, line 175: Pool size here should just be 2 (1 for analyzer, 1 for finder). There was an error in previous versions of the code, so revert this change.

dwalluck:
@michalszynkiewicz says that he does not need Pnc support when finding builds. In fact, Pnc support will not give a valid nvr (name-version-release) list. Why? Build Finder tries to provide a single view `KojiBuild` for both Koji and Pnc builds. Therefore, it gives Pnc builds an nvr, and while the version is (usually) correct, the release is always set to 1 since Pnc has no similar concept. Besides, these builds will not exist in Koji regardless of this information.

@matejonnet We need to update the included `config.json` in the source tree to set `buildSystems` to `[ "koji" ]`. I am not sure if we should include a full version (minus internal URLs, of course) or leave the minimal version. With the minimal configuration, the defaults do not become clear and we may want to change some other defaults as well.

dwalluck:
@matejonnet I was speaking with @thescouser89 and I realized that the dependencies should be `core` and `report` and not `build-finder`. The `build-finder` dependency is the standalone cli jar. Either way, we may have some issue with `logback.xml` which we will fix in the next build finder snapshot.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
matejonnet: Well spotted, sorry that got in with a merge conflict.

matejonnet:
> @matejonnet We need to update the included `config.json` in the source tree to set `buildSystems` to `[ "koji" ]`. I am not sure if we should include a full version (minus internal URLs, of course) or leave the minimal version. With the minimal configuration, the defaults do not become clear and we may want to change some other defaults as well.

Added the config you mentioned, about other configs I'm relay not familiar with the build finder, please let me know if some other config is required.

matejonnet:
Re-test this please!

thescouser89:
retest this please

PR #378, opened 2020-08-26
Fix setting of file mode and last modified time

dwalluck:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileUtils.java
```diff
@@ -107,14 +111,103 @@ public static String getCompressorType(final String filename) {
         }
     }
 
+    public static void setModeAndLastModifiedTime(Path path, ArchiveEntry entry) throws IOException {
```
michalszynkiewicz, line 114: I vaguely remember we had such code in pig and we had to drop it because it caused problems in some environments. Unfortunately I don't remember where was it. @Ladicek @paulgallagher75 maybe you guys do?

@dwalluck what gain do you see in preserving these attributes?

dwalluck: Well, 3 out of the 4 methods already had it. Do you recall dropping it from just untar I think it was? It's supposed to behave like native tar as much as possible, and it would preserve them, so it shouldn't be causing problems. The issue is for example it won't be preserving an executable script inside the archive.

dwalluck: Mostly what I did was factor the common code that was already there out into separate methods. If we need to, we could skip it in untar or whatever was a problem but where is the issue (I hope one was filed for this).

michalszynkiewicz: I found the commit that reverted the thing in pig and it turns out there the code was also setting the owner. Maybe that was the problemmatic part...
Let's merge it and roll back if it causes problems.

dwalluck: We _could_ set the owner, but I don't think we want to store that inside. For tar it's useful when you're doing backups.

PR #381, opened 2020-08-26
All methods in CLIExecutor can be static

dwalluck:


janinko:
Why is that better?

rnc:
Generally most of the sub-command classes could be static but until NCL-5974 is resolved I am not sure its worth the time changing them

dwalluck:
Well, I should say that it's already been written as a utility class. It was just missing `static` on some of the already (should be) static methods, meaning that it doesn't use any instance variables or state.

Maybe this is related to what @rnc is saying, but we also have these `Utils` classes, but there is some inconsistency there. Generally, a class like this (with only `static` methods) is marked `final` and has a `private` constructor. Some in the codebase do, and some don't.

rnc:
Add `@UtilityClass` annotation from lombok?

dwalluck:
> Generally most of the sub-command classes could be static but until NCL-5974 is resolved I am not sure its worth the time changing them

Also, I believe the nested inner classes in the command classes are supposed to be `static`. IDEA complains that not doing so is also a security risk. Some information about that here https://stackoverflow.com/questions/53847742/why-compile-non-static-inner-class-will-creation-of-hidden-package-visible-meth.

dwalluck:
> Add `@UtilityClass` annotation from lombok?

Not a bad idea, IMO.

rnc:
> > Generally most of the sub-command classes could be static but until NCL-5974 is resolved I am not sure its worth the time changing them
> 
> Also, I believe the nested inner classes in the command classes are supposed to be `static`. IDEA complains that not doing so is also a security risk. 

Likely not an issue for Bacon but is blocked by 5974 regardless IMO.

dwalluck:
> Likely not an issue for Bacon but is blocked by 5974 regardless IMO.

I was talking more about the `Utils` classes not the `Cli` classes in this case. It seems that a drawback to lombok's `@UtilityClass` is that you can't use static imports from that class (as the compiler does not see them). Now, I don't tend to use static imports much, but they're used quite a bit in the bacon code. I made an attempt to use this annotation, but it may not be worth it.

dwalluck:
This will be superseded by `@UtilityClass` PR which I plan to submit soon, although even I am not sure if that annotation is worth the trouble.

dwalluck:
@rnc Attempt at #422.

thescouser89:
picocli PR merged!

rnc:
@dwalluck  Given #422 has conflicts should this one be merged and call that a day on the static conversions?

dwalluck:
> @dwalluck Given #422 has conflicts should this one be merged and call that a day on the static conversions?

Sure, although no one seems to have actively approved this one either.

rnc:
@thescouser89 Can this be either merged or reviewed please?

PR #386, opened 2020-08-27
Remove use of string concatenation with loggers

dwalluck:


pnc/src/main/java/org/jboss/pnc/bacon/pnc/client/BifrostClient.java
```diff
@@ -40,7 +40,7 @@ public void writeLog(String id, boolean follow, Consumer<String> onLine) throws
         }
 
         URI logsUrl = baseUrl.resolve(URI.create("/text?" + query));
-        logger.debug("Reading logs from {}.", logsUrl.toString());
+        logger.debug("Reading logs from {}", logsUrl);
```
michalszynkiewicz, line 43: `logger` has to be changed to `log` here, doesn't it?

dwalluck: @michalszynkiewicz Fixed

PR #387, opened 2020-08-27
Minor javadoc fixes

dwalluck:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/SleepUtils.java
```diff
@@ -52,7 +52,7 @@ public static void waitFor(Supplier<Boolean> condition, int checkInterval, boole
     }
 
     /**
-     * Wait until the <code>condition</code> returns a non null value or time out is reached
+     * Wait until the {@code condition} returns a non null value or time out is reached
```
michalszynkiewicz, line 55: what's the difference?

michalszynkiewicz (approved):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/SleepUtils.java
dwalluck: Since `{@code}` is a javadoc tag, it allows javadoc to format code the way it wants (and I suppose it's a good idea not to mix `{@code}` and `<code></code>`.

The `{@literal}` tag denotes text which does not contain HTML markup or nested javadoc tags.
The `{@code}` tag is equivalent to <code>{@literal abc}</code>.

And so actually some of the changes above could have used `{@literal}` instead of HTML markup.

michalszynkiewicz: Thanks!

PR #396, opened 2020-08-28
Clean context on config change

michalszynkiewicz:


michalszynkiewicz:
@paulgallagher75 with it when you modify the config yaml the context will get cleared. Also `pig run` and `pig configure` will always start clean

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/HashUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jboss.resteasy.util.Hex;
+
+public class HashUtils {
+    private HashUtils() {
+    }
+
+    public static String hashDirectory(Path directory) {
+        MessageDigest sha = DigestUtils.getSha512Digest();
+        try {
+            Files.walk(directory).filter(Files::isRegularFile).sorted().forEach(path -> {
```
dwalluck, line 18: I think this should be `try (FIles.walk()) {`

dwalluck: I'll fix it in #402

michalszynkiewicz: thanks!

PR #397, opened 2020-08-28
[NCL-5927] Add --no-external-scm flag for scm-repository update

thescouser89:
This will allow the user to say we don't want an external-scm for that
scm repository.

If both the '--no-external-scm' and '--external-scm <value>' are
specified, then a validation fails and prevents the user to proceed.

The '--no-external-scm' automatically sets the preBuild to false.


dwalluck:
@thescouser89 picocli has negatable options https://picocli.info/#_negatable_options. Does aesh not have this? Having to implement two different flags for each boolean option doesn't sound like fun.

rnc:
> @thescouser89 picocli has negatable options https://picocli.info/#_negatable_options. Does aesh not have this? Having to implement two different flags for each boolean option doesn't sound like fun.

+1, I would vote we avoid workarounds like this as it adds evidence to consider switching to picocli. I'd be interested to try to work on that

thescouser89:
No such thing in aesh unfortunately.

PR #401, opened 2020-08-28
Use entrySet rather than get when walking through map

dwalluck:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
@@ -92,15 +92,16 @@ public class RepositoryUtils {
         for (String path : pomPaths) {
             ArtifactPathInfo artifactPathInfo = ArtifactPathInfo.parse(path);
             ProjectVersionRef gav = artifactPathInfo.getProjectId();
-            List<SingleVersion> singleVersions = new ArrayList<SingleVersion>();
+            List<SingleVersion> singleVersions = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
```diff
@@ -92,15 +92,17 @@ private static void generateMetadata(File mavenRepositoryDirectory, Set<String>
...
+            List<SingleVersion> singleVersions = new ArrayList<>();
             if (!metas.isEmpty() && metas.containsKey(gav.asProjectRef())) {
                 singleVersions = metas.get(gav.asProjectRef());
             }
             singleVersions.add((SingleVersion) gav.getVersionSpec());
             metas.put(gav.asProjectRef(), singleVersions);
         }
-        for (ProjectRef ga : metas.keySet()) {
-            List<SingleVersion> singleVersions = metas.get(ga);
+        Set<Map.Entry<ProjectRef, List<SingleVersion>>> entries = metas.entrySet();
```
michalszynkiewicz, line 102: The code is a lot longer now, could you inline metas.entrySet() to make it a bit shorter?

dwalluck: Inside the `for()`? I wasn't sure if that was a performance issue.

dwalluck: OK, this just increases it 1 line compared to the previous code.

PR #402, opened 2020-08-28
Close streams for Files.walk/list, GenericSettingClient, and CSV

dwalluck:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ParentPomDownloader.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ParentPomDownloader.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Stream;
 
 /**
  * @author Ken Finnigan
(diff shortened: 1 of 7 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/CSVUtils.java
```diff
@@ -1,25 +1,25 @@
...
-        return CSVFormat.EXCEL.withDelimiter(delimiter)
-                .withFirstRecordAsHeader()
-                .parse(in)
-                .getRecords()
-                .stream()
-                .map(record -> record.get(columnName))
-                .collect(Collectors.toSet());
+        try (Reader in = Files.newBufferedReader(Paths.get(filePath), StandardCharsets.UTF_8);
+                CSVParser parser = CSVFormat.EXCEL.withDelimiter(delimiter).withFirstRecordAsHeader().parse(in)) {
+            return parser.getRecords().stream().map(record -> record.get(columnName)).collect(Collectors.toSet());
```
michalszynkiewicz, line 21: @thescouser89 could we do something so that the autoformatter doesn't squash all the calls to a single line?

dwalluck: My problem is with the `try`. Why is indenting like that when it breaks the two statements? I wish it would line it up all statements with the `try(`. It makes it really hard to read, IMO.

I thought that with a chain of calls like this, it already does align methods (on the dot `.`) if there's a certain number (or length). But again, I do not like how it aligns all methods except the first line which could have multiple method calls and is not aligned with the rest.

My only gripes (I think) with the current style is how it breaks lines on a open parenthesis and how it align lines. I think I was outvoted on the first, but the second issue may apply to that as well. Also, some cases may be unfixable due to how the formatter works.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ParentPomDownloader.java
```diff
@@ -56,12 +57,13 @@ private void process() throws IOException {
...
 
     private void doProcess() throws IOException {
-        Files.walk(repoPath)
-                .filter(Files::isRegularFile)
-                .filter(ParentPomDownloader::isPom)
-                .map(Pom::new)
-                .filter(Pom::hasParent)
-                .forEach(p -> processPom(repoPath, p));
+        try (Stream<Path> stream = Files.walk(repoPath)) {
+            stream.filter(Files::isRegularFile)
```
dwalluck, line 61: @michalszynkiewicz @thescouser89 See, it aligned this one, but why doesn't it align it with the first dot? And why didn't it align the other statement?

dwalluck: Because it does it only if it exceeds the max line length, not always.

PR #403, opened 2020-08-28
change nvr list generator to use paths

michalszynkiewicz:
### Checklist:

* [hell no!] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) for your change if user-facing?
* [nope] Have you added unit tests for your change?

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
@@ -17,8 +17,6 @@
  */
 package org.jboss.pnc.bacon.pig.impl.nvr;
 
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.FilenameUtils;
 import org.jboss.pnc.bacon.pig.impl.utils.BuildFinderUtils;
 import org.jboss.pnc.build.finder.koji.KojiBuild;
 import org.jboss.pnc.build.finder.report.NVRReport;
@@ -28,6 +26,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
```diff
@@ -38,11 +39,11 @@
     private NvrListGenerator() {
     }
 
-    public static boolean generateNvrList(Map<String, Collection<String>> checksums, String targetPath) {
+    public static boolean generateNvrList(Map<String, Collection<String>> checksums, Path targetPath) {
         log.info("Generating NVR list for {} checksums and saving result to {}", checksums.size(), targetPath);
 
         List<KojiBuild> builds = BuildFinderUtils.findBuilds(checksums, true);
-        File outputDirectory = new File(FilenameUtils.getPath(targetPath));
+        File outputDirectory = targetPath.toFile().getParentFile();
```
dwalluck, line 46: Might be better to get the parent and then do the conversion `targetPath.getParent().toFile()`?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
```diff
@@ -54,10 +55,16 @@ public static boolean generateNvrList(Map<String, Collection<String>> checksums,
...
 
+        if (destFile.isFile()) {
+            if (!destFile.delete()) {
+                throw new RuntimeException(
+                        "Failed to delete the previously created nvr list: " + destFile.getAbsolutePath());
+            }
+        }
         try {
-            FileUtils.copyFile(srcFile, destFile, true);
+            FileUtils.moveFile(srcFile, destFile);
```
dwalluck, line 67: OK, this is why I changed it to `copyFile()` originally. I was having an issue with the `destFile` already existing, I think. As long as we're sure we actually want to move the file (rerunning should overwrite this file anyway)/

pom.xml
```diff
@@ -18,14 +18,15 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
dwalluck, line 21: This file may be better as a separate commit.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
michalszynkiewicz: With copying, the problem was that when the target directory was fixed, the nvr.txt was created there and polluted the release directory. 
Alternatively to moving we can create the file in some temp location and copy it but copying is a bit easier.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
michalszynkiewicz: do you think there is any real difference between the two?

pom.xml
michalszynkiewicz: sorry for it, I'll do better next time :)

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
dwalluck: There were two nvr txt files in the directory I think? I am sorry that the Build Finder API just doesn't allow you to change the output file name in the first place.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
dwalluck: My original idea was just to use the new `Path` API and save the conversion to `File` to the last moment.

With `Path.getParent()`, it is going to give weird behavior if the path is not `normalize()`d first. I am not sure if the `File` API does the same thing. For example, I think the parent of `parent/dir/../` is `parent/dir` and not `parent` like you would want.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
dwalluck: Actually, you can call `report.setBaseFilename()` to set the name. It will automatically add the`.txt` extension though. Doesn't it make sense to write it directly? Then we can remove the call to `moveFile()` altogether.

Here's what it calls internally:

```
FileUtils.writeStringToFile(new File(this.outputDirectory, this.baseFilename + ".txt"), text, StandardCharsets.UTF_8, false);
```

michalszynkiewicz: that simplifies things a lot!

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/nvr/NvrListGenerator.java
```diff
@@ -38,31 +37,31 @@
...
-        File destFile = new File(targetPath);
+        return true;
+    }
 
-        try {
-            FileUtils.copyFile(srcFile, destFile, true);
-        } catch (IOException e) {
-            log.error("Failed to copy file {} to {}: {}", srcFile, destFile, e.getMessage(), e);
-            return false;
+    private static String baseFileName(Path targetPath) {
```
dwalluck, line 58: There is also https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/FilenameUtils.html#getBaseName-java.lang.String- whichamke make your code smaller.

PR #412, opened 2020-09-02
Add impsort-maven-plugin

dwalluck:
Add impsort-maven-plugin, by the authors of formatter-maven-plugin, for sorting imports. This helps keep imports consistent between different developers and IDEs. I tried to choose the options closest to what was already being used in the code (the ones that caused the smallest diff).


rnc:
Good idea. Do we have a suitable configuration for IntelliJ so it doesn't keep having to reload the files after a command line build which imposes file changes?

dwalluck:
I believe that the only difference between the current settings and the default IDEA order is that there is a space between between `javax.` and `java.`  (and I don't see an option to not include a space when you define a group).

```
<code_scheme name="Default" version="173">
  <JavaCodeStyleSettings>
    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
    <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
      <value />
    </option>
    <option name="IMPORT_LAYOUT_TABLE">
      <value>
        <package name="" withSubpackages="true" static="false" />
        <emptyLine />
        <package name="javax" withSubpackages="true" static="false" />
        <emptyLine />
        <package name="java" withSubpackages="true" static="false" />
        <emptyLine />
        <package name="" withSubpackages="true" static="true" />
      </value>
    </option>
  </JavaCodeStyleSettings>
</code_scheme>
```

PR #414, opened 2020-09-02
Remove hardcoded line separator from Shell.writeln

dwalluck:
I am not sure if you wanted an extra newline there to begin with, but try letting shell handle `writeln()` in case it is system-dependent. If we want another newline, we can try `shell.writeln("").`


thescouser89:
I actually wanted the newline David to provide better separation between 'Examples' and the actual example.

dwalluck:
> I actually wanted the newline David to provide better separation between 'Examples' and the actual example.

OK, I added `shell.writeln("");`, if that works. Oddly, there is no `shell.writeln();`

thescouser89 (approved):

PR #415, opened 2020-09-02
Make test classes and methods package-private as per JUnit 5

dwalluck:
JUnit 4 forced test classes and methods to be public. JUnit 5 changed this. See also <https://github.com/junit-team/junit5/issues/679> where they made the change to their own codebase. This is a minor PR, but it adds some consistency (some tests were using this and some weren't).

### DIFF
--- a/integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
+++ b/integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
@@ -34,27 +34,27 @@ public class AbstractTest {
     protected PNCWiremockHelper wmock = new PNCWiremockHelper();
 
     @BeforeAll
-    public static void startWiremockServer() {
+    static void startWiremockServer() {
         wireMockServer = new WireMockServer(options().port(8080).notifier(new ConsoleNotifier(false)));
         wireMockServer.start();
     }
 
     @AfterAll
-    public static void stopWiremockServer() {
+    static void stopWiremockServer() {
         wireMockServer.stop();
     }
 
     @BeforeEach
-    public void stubBabnner() {
+    void stubBanner() {
         wmock.init();
     }
 
     @AfterAll
-    public void clearStubs() {
+    static void clearStubs() {
         wireMockServer.resetAll();
     }
 
-    public static String getRandomString() {
+    protected static String getRandomString() {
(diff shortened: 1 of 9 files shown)


rnc:
What happens if someone uses Jupiter Junit4 test methods - won't they still need to be public? (e.g. for using older powermock extensions). Otherwise lgtm.

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
```diff
@@ -34,27 +34,27 @@
...
 
     @BeforeEach
-    public void stubBabnner() {
+    void stubBanner() {
         wmock.init();
     }
 
     @AfterAll
-    public void clearStubs() {
+    static void clearStubs() {
```
dwalluck, line 53: Is one of these supposed to be `@AfterEach` (there are two `@AfterAll` here)?

dwalluck:
> What happens if someone uses Jupiter Junit4 test methods - won't they still need to be public? (e.g. for using older powermock extensions). Otherwise lgtm.

Do you mean the `junit-vintage-engine`? It doesn't seem to need it (although I thought this was justa  migration aid): https://github.com/junit-team/junit5-samples/tree/main/junit5-migration-maven/src/test/java/com/example/project

dwalluck:
As an aside, I think mockito is now supporting mocking `private` and `final` methods which obviates a lot of the need for powermock, and I think the mockito team itself is managing a jupiter extension. But I don't know about the powermock extension specifically.

rnc (approved):

dwalluck:
Also, docs are saying that `@AfterAll` and `@BeforeAll` methods must be static, but the annotation doesn't enforce this, and you just seem to get weird behavior, like the method won't run but the compiler doesn't complain at all.

dwalluck:
".. methods must have a void return type, must not be private , and must be static" but it seems to allow you to annotate without it being such, leading to issues?

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
thescouser89: Looks like it should be AfterEach right? It does feel like that would be the right one.

PR #420, opened 2020-09-03
[NCL-6018] Download logs even when build successful

thescouser89:
In a previous commit, we didn't download logs if the builds were
successful because the logs would be printed in the stdout and would be
way too long.

However this broke addons that rely on the logs. This commit reverts the
changes of the previous commit, and also doesn't print the logs
(successful or not) to the stdout.

The addons broken are: MicroProfileSmallRyeCommunityDepAnalyzer.java and
RuntimeDependenciesAnalyzer.java

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/BuildInfoCollector.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/BuildInfoCollector.java
@@ -105,17 +105,15 @@ public class BuildInfoCollector {
 
             PncBuild result = new PncBuild(build);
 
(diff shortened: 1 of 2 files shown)


michalszynkiewicz:
@thescouser89 could you make the add-ons download the build logs if they are not there instead of downloading them even if not needed? Only if it's a trivial change

thescouser89:
@michalszynkiewicz  I could perhaps download the logs on the 'getBuildLogs' method call (and cache the results) and remove the 'setBuildLogs' on PncBuild. WDYT?

michalszynkiewicz:
If that's easy, :+1:

PR #422, opened 2020-09-03
Use lombok's UtilityClass on utility classes

dwalluck:
### For

* Class marked `static final` and all fields automatically marked `static`
* Private constructor created which throws `IllegalArgumentException`

### Against

* Can't be used in cases where we have an inner class that uses a public constructor, but maybe this isn't a great design to begin with
* Can't be used with `import static`. Once you use `@UtilityClass`, the compiler does not know that these fields are `static`.

### DIFF
--- a/common/src/main/java/org/jboss/pnc/bacon/common/Constant.java
+++ b/common/src/main/java/org/jboss/pnc/bacon/common/Constant.java
@@ -1,20 +1,19 @@
 package org.jboss.pnc.bacon.common;
 
+import lombok.experimental.UtilityClass;
+
 import java.io.File;
 
+@UtilityClass
 public class Constant {
-
-    public static final String DEFAULT_CONFIG_FOLDER = System.getProperty("user.home") + File.separator + ".config"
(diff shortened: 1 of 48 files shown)


michalszynkiewicz:
I don't see much benefit in terms of code size, it makes what's going on less clear. Sorry, I'm a :-1: on it

dwalluck:
> I don't see much benefit in terms of code size, it makes what's going on less clear. Sorry, I'm a 👎 on it

Then we likely shouldn't use it, but some classes should probably be updated manually. I will request some reviews to get some other opinions before closing, particular @rnc because he suggested it. Also, @michalszynkiewicz started neutral on it.

michalszynkiewicz:
yes, I didn't know how `@Utility` works in practice then, sorry.

rnc:
I am neutral on this, and I only mentioned it in passing on #381. I am not completely convinced its necessary apart from removing some IDE warnings? Given I changed a lot of the code when converting to picocli (hence the conflicts now) perhaps this should be dropped?

PR #423, opened 2020-09-04
Add license headers and use license-maven-plugin to check for them

dwalluck:
### Checklist:


michalszynkiewicz:
We don't do it in Quarkus, so apparently this is not a requirement.
If not required, I'd rather not have licenses in each file.
We have a LICENSE file in the root directory and a proper license definition in the pom.

I'm closing the PR but please reopen if there's a valid explanation to have it.

rnc:
I believe the licensing recommendations have changed so I don't think this is necessary either

PR #424, opened 2020-09-04
Format files and use editorconfig-maven-plugin to check them

dwalluck:
### Checklist:


.editorconfig
```diff
@@ -9,3 +9,12 @@ insert_final_newline = true
 max_line_length = 120
 tab_width = 4
 trim_trailing_whitespace = true
+
+[eclipse-codeStyle.xml]
+indent_size = unset
+indent_style = space
+
+[{integration-tests/src/test/resources/logback-test.xml,pig/src/main/resources/*.xml,pig/src/test/resources/*.xml}]
```
dwalluck, line 17: @thescouser89  I know you prefer 4 spaces. Should these files just be converted to use 4 (instead of 2) spaces? The `eclipse-codeStyle.xml` is machine sgenerated and seems unindented, so it may be better to leave that one excluded.

pom.xml
```diff
@@ -535,6 +535,24 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.ec4j.maven</groupId>
+                <artifactId>editorconfig-maven-plugin</artifactId>
+                <version>0.1.0</version>
+                <configuration>
+                    <excludes>
```
dwalluck, line 543: Excluded `.bacon/` just in case it's run inside the tee. Excluded `.cache/` for formatter-maven-plugin.

michalszynkiewicz: is it the same plugin that's used on Jenkins?

dwalluck: I don't know anything about Jenkins. This plugin was made by someone who works for Red Hat/Quarkus and I have contributed a few PRs to it.

michalszynkiewicz: What is the benefit of adding it?

rnc:
I think having a developer guide with correct setup instructions for an IDE to avoid constant reloading when command line builds reformat the files would be more helpful.

dwalluck:
> I think having a developer guide with correct setup instructions for an IDE to avoid constant reloading when command line builds reformat the files would be more helpful.

Actually, the IDEA support for `.editorconfig` is better as I believe it autoloads it with no manual work required. And if PiG prefers different formatting for XML files, another `.editorconfig` should probably be placed in the `pig` directory. Right now, it is just set to `check` and not `format`. Another option is to remove the executor and just allow manual invocation.

For the import order, I exported an XML file, but I don't know where to place it in the tree/git. I may be wrong, but I don't think it can be placed in `.idea/` and autoloaded, which is too bad.

pom.xml
dwalluck: It is similar to the formatter-maven-plugin, but at a lower level, and supports other file types besides Java. On Windows, for example, the default IDEA settings are not good. IDEA would allow the wrong EOLs to be created and, in fact, changes every file you edit. Even on Unix, soem editors don't seem to add a newline at end of file.

The `.editorconfig` file automatically tells editors like IDEA to use Unix newlines (and some other settings you can see in that file). The plugin can check if files created outside of an IDE adhere to the format used in the project.

.editorconfig
thescouser89: I do prefer 4 spaces.

Is that an ignore list? If so it looks fine.

rnc: I need a wider monitor!!

dwalluck: > I do prefer 4 spaces.
> 
> Is that an ignore list? If so it looks fine.

These patterns are set to 2 spaces instead of 4 (4 is the global option).

There are two options: leave integration-tests and pig using 2 and the rest use 4, or move everything to 4. If they're going to use 2, rather than add to the global `.editorconfig` like it is here, it may be better to place a new `.editorconfig` in certain directories with `*.xml`. This would allow pig to have different global formatting from the rest of bacon if it wants.

dwalluck: > I need a wider monitor!!

You mean this line? There is no way to break it but you can come up with a more terse regex, or just use a separate file.

Too bad they basically got rid of 16:10 monitors, but they do have the 21:9 ultrawides now!

rnc: > > I need a wider monitor!!
> 
> You mean this line? There is no way to break it but you can come up with a more terse regex, or just use a separate file.
> 
> Too bad they basically got rid of 16:10 monitors, but they do have the 21:9 ultrawides now!

Er, no,  sorry that was referring to the 4 spaces!

rnc: > > I do prefer 4 spaces.
> > Is that an ignore list? If so it looks fine.
> 
> These patterns are set to 2 spaces instead of 4 (4 is the global option).
> 
> There are two options: leave integration-tests and pig using 2 and the rest use 4, or move everything to 4. If they're going to use 2, rather than add to the global `.editorconfig` like it is here, it may be better to place a new `.editorconfig` in certain directories with `*.xml`. This would allow pig to have different global formatting from the rest of bacon if it wants.

Why are some parts using different formatting?

dwalluck: I assume because bacon and pig were two separate codebases and then became one.

rnc: I think the formatting in the codebase should be consistent and changed.

pom.xml
rnc: I am not completely convinced here - there appears to be some overlap between the two plugins which would increase the maven build time. Given currently the primary platform for Bacon is Linux, what other benefits does adding this plugin give?

thescouser89:
I'm inclined to approve this PR. Like David said, Intellij does support .editorconfig (https://plugins.jetbrains.com/plugin/7294-editorconfig) and that would be exactly the IDE support we'd need.

I'm against having 2 .editorconfig for PNC and PiG. It makes sense to have 1 unified one, whatever it is. If PiG has a strong preference for 2 spaces for xml, so be it :) I don't care that much.

thescouser89 (approved):

dwalluck:
> I'm against having 2 .editorconfig for PNC and PiG. It makes sense to have 1 unified one, whatever it is. If PiG has a strong preference for 2 spaces for xml, so be it :) I don't care that much.

So, I can update the PR to reformat those files and remove the exclusions.

The only one I wasn't sure about was the eclipse one as that is the way eclipse outputs it. After reading the editorconfig docs, I am still not sure if that is the correct way to have 0 indent level. It seems like it doesn't explicitly support setting it to 0.

rnc:
> I'm inclined to approve this PR. Like David said, Intellij does support .editorconfig ([plugins.jetbrains.com/plugin/7294-editorconfig](https://plugins.jetbrains.com/plugin/7294-editorconfig)) and that would be exactly the IDE support we'd need.
> 
> I'm against having 2 .editorconfig for PNC and PiG. It makes sense to have 1 unified one, whatever it is. If PiG has a strong preference for 2 spaces for xml, so be it :) I don't care that much.

I'm still not convinced and am currently a -1 on this PR. I do *not* see a need for another formatter plugin right now - its not clear to me what the benefit is.

This conversation has raised up a few points:
1. The inconsistent project formatting should be solved (in a separate PR IMHO).
2. `.editorconfig` support is implicit within IntelliJ and no plugin is required for that as far as I know.

rnc (changes requested):
Currently -1 - see my last comment.

dwalluck:
> 2\. `.editorconfig` support is implicit within IntelliJ and no plugin is required for that as far as I know.

I checked and the EditorConfig plugin is by JetBrains itself and is listed as "bundled".

.editorconfig
dwalluck: OK, so running `editorconfig:format` automatically fixes the formatting.

pom.xml
dwalluck: I tested build time with and without the plugin. I didn't even see a consistent increase in build time taken when it was enabled, sometimes the build was even faster.

This ensures that files are formatted correctly, regardless of IDE used. The indent level of XML files is not dependent on OS at all.

Moreover, I see, e.g., in the pnc codebase with no such editconfig files that someone's editor is creating files without a newline, which is a violation even on Linux. This is required by POSIX https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_403.

.editorconfig
rnc: Ok.

pom.xml
rnc: Ok.

rnc (approved):

dwalluck:
I've put the file reformatting on this same PR because after removing the exclusions and making all files have the same format, the check will fail, and @thescouser89 I believe wanted `check` changed to `format`. It serves as a good example of what the plugin can do.

I also don't know if people are getting the same behavior in terms of the build time, but I didn't notice much change. It's possible this could be moved to a profile. Our releng parent had a profile, but I ended up not liking it too much because a lot of issue would pop up at the last moment at release time instead of PR time.

One issue is that you need to add exclusions or be careful what you have in the source tree. For example, I think I was creating build.log in the tree, then it might try to do something with this file.

PR #431, opened 2020-09-07
NCLl5974 - Replace aesh with picocli

rnc:
### Checklist:


rnc:
@michalszynkiewicz  @thescouser89  for early feedback :-)

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -17,156 +17,190 @@
...
 
 /**
  * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com <br>
  *         Date: 12/13/18
  */
 
 @Slf4j
-@GroupCommandDefinition(name = "bacon", description = "Bacon CLI", groupCommands = { Pnc.class, Da.class, Pig.class })
-public class App extends AbstractCommand {
+@Command(name = "bacon.jar", versionProvider = VersionProvider.class, subcommands = { Da.class, Pig.class, Pnc.class })
```
michalszynkiewicz, line 64: why .jar in the name?

rnc: Er, no reason :-) It does mean I can remove `commandLine.setCommandName("bacon");`. Will push a change.

dwalluck: I guess the "official" way to run it is via the `bacon` script as opposed to `java -jar bacon.jar`, therefore, I think bacon for the same makes sense.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -17,156 +17,189 @@
...
-                                    "[bacon@console]$ ",
-                                    new TerminalColor(Color.DEFAULT, Color.DEFAULT, Color.Intensity.BRIGHT))));
-
-            console.start();
-
-            return 0;
-        } else {
-            registry = AeshCommandRegistryBuilder.builder().command(this.getClass()).create();
+            // From https://github.com/remkop/picocli/wiki/JLine-3-Examples
+            AnsiConsole.systemInstall();
```
dwalluck, line 126: I used this line in build finder, but I couldn't figure out how to guarantee how `AnsiConsole.systemUninstall()` would be called, so I put the install in a `static` block and uninstall in a shutdown handler. You _must_ call uninstall once for every install or there is some resource leak.

But, now I think this is actually only used on old Windows versions (< 10), so it may be able to be left out entirely. You can even see in the picocli code logic like

```
if (isWindows && !Ansi.AUTO.enabled()) {
    AnsiConsole.systemInstall();
}
```

rnc: I guess I could put the uninstall in the EndOfFileException block (when ctrl-d) is called which exits the program. However I can't see where your code is sourced from - I used https://github.com/remkop/picocli/tree/master/picocli-shell-jline3

rnc:
retest this please

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
dwalluck: I may have overexaggerated the need for it---it looks like uninstall just resets the I/O streams, but it still looks like a bug in that jline3 example.

The one I was talking about is from here: https://github.com/remkop/picocli/blob/34a7d3ab72563678303dd02e683f1ad599732c49/src/test/java/picocli/HelpAnsiTest.java#L118 and Jansi's own examples: https://github.com/fusesource/jansi/blob/146109610a85892efdad4f50a1a974d9d6ac21ee/example/src/main/java/org/fusesource/jansi/Main.java#L10

rnc: I have added an uninstall in the latest version for when the user quits the console. Perhaps you could enter an issue in picocli jline3 - the author is very responsive.

dwalluck: I'll file an issue against picocli. Maybe he knows whether it is necessary, too.

dwalluck: He replied to https://github.com/remkop/picocli/issues/1165 and added the uninstall to the docs. It also seems that this could be made conditional or even removed. See what you think after you read the discussion on that issue.

thescouser89:
First, wow that's a huge effort and I can't figure out how you managed to do it in such little time. I truly don't know how you did it!!! That's outstanding work!

The outputs look nicer and more descriptive.

I just noticed a tiny bug on the pig side, where the 'tempBuild' seems to be listed as a command rather than a flag?

```
/tmp/bacon NCL5974 ❯ java -jar cli/target/bacon.jar pig configure -h
Usage: bacon pig configure [-hov] [--clean] [--skipBranchCheck] [t] [-p=<configurationFileLocation>] [--profile=<profile>] [--releaseStorageUrl=<releaseStorageUrl>] <configDir>
Configure PNC entities
      <configDir>           Directory containing the Pig configuration file
      --clean               If enabled, the pig execution will not attempt to continue the previous execution
  -h, --help                display this help message
  -o                        use json for output (default to yaml)
  -p=<configurationFileLocation>
                            Path to PNC configuration folder
      --profile=<profile>   PNC Configuration profile
      --releaseStorageUrl=<releaseStorageUrl>
                            Location of the release storage, typically on rcm-guest staging. Required for auto-incremented milestones, e.g. DR*
      --skipBranchCheck     If set to true, pig won't try to determine if the branch that is used to build from is modified. Branch modification check takes a lot of time, if you use tag,
                              this switch can speed up the build.
      t, tempBuild          If specified, PNC will perform temporary builds
  -v                        Verbose output
/tmp/bacon NCL5974 ❯ 
```

The profile switch also doesn't seem to be working:
```
/tmp/bacon NCL5974 ❯ java -jar cli/target/bacon.jar pnc whoami --profile prod                                                                                                              
Enter your password: 
---
id: "102"
username: "dcheung"

/tmp/bacon NCL5974 ❯ java -jar cli/target/bacon.jar pnc whoami --profile stage                                                                                                              
---
id: "102"
username: "dcheung"
```

In aesh bacon:
```
~/projects/work/bacon nclsup-119 ❯ java -jar cli/target/bacon.jar pnc whoami --profile prod
Enter your password: 
---
id: "2"
username: "dcheung"

~/projects/work/bacon nclsup-119 ❯ java -jar cli/target/bacon.jar pnc whoami --profile stage                                                                                                6s
Enter your password: 
---
id: "1"
username: "dcheung"
```

It only seems to load the default profile.

The '--verbose' flag is also missing (only `-v` present). Same for '--configPath'.

That's all I could really find. It's great work! :)

rnc:
@thescouser89 Thanks for the kind review and finding those things! I've pushed some fixes and some more tests to cover them.

michalszynkiewicz:
:+1: from me, would like @thescouser89 to approve since he did the work on aesh

thescouser89:
LGTM with the latest changes. Merging! :+1:

PR #433, opened 2020-09-07
[NCL-6018] Split log lines across multiple items

brusdev:
Split log lines across multiple items to fix NotYetAlignedFromDependencyTree.

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuild.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuild.java
@@ -88,6 +88,10 @@ public class PncBuild {
         buildLog = log == null ? Collections.emptyList() : asList(log.split("\\r?\\n"));
     }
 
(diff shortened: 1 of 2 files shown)


thescouser89:
LGTM!

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuild.java
```diff
@@ -88,6 +88,10 @@ public void addBuildLog(String log) {
         buildLog = log == null ? Collections.emptyList() : asList(log.split("\\r?\\n"));
```
michalszynkiewicz, line 88: is this method still used?

brusdev: No, ATM it isn't used. Can I remove it or you or do you prefer to keep it for future use?

rnc:
Should this PR not have a test associated with it?

PR #436, opened 2020-09-08
[NCLSUP-119] Fix check for isBranchModified

thescouser89:
The check assumed that the 'scmRevision' in the build information would
be a tag name. Instead it is the commit id.

This commit fixes it by grabbing the scmTag information instead which
contains the real tag name.

It also treats the 'branch' as either a ref (pointer to a commit), or
the commit itself. This should handle the corner case where the full commit
id is specified in the build-config.yaml

Optionals are also now used to better handle the case where the
information is not available

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/GitRepoInspector.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/GitRepoInspector.java
@@ -34,8 +34,10 @@ import java.io.File;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


thescouser89:
Example of a run:

```
...
[INFO ] - Adding/updating build configurations
[INFO ] - Trying to check if branch 'fancy-branch' in 'git+ssh://{git server}/kiegroup/lienzo-core.git' has been modified, compared to latest build of build config '3456'
[WARN ] - close() called when useCnt is already zero for Repository[/tmp/git7862001070209087875/.git]
[INFO ] - Trying to check if branch 'fancy-branch' in 'git+ssh://{git server}/kiegroup/lienzo-tests.git' has been modified, compared to latest build of build config '2345
[WARN ] - close() called when useCnt is already zero for Repository[/tmp/git1932688888194040680/.git]
[INFO ] - Trying to check if branch 'fancy-branch' in 'git+ssh://{git server}/prod-izpack.git' has been modified, compared to latest build of build config '1234'
[WARN ] - close() called when useCnt is already zero for Repository[/tmp/git8402590848080845724/.git]
[INFO ] - Temporary build
[INFO ] - Temporary build
[INFO ] - Performing builds of build group 123 in PNC
[INFO ] - Waiting for finish of group build 456
...
```

rnc:
Should we add a test here?

PR #437, opened 2020-09-09
[NCL-6028] properly parallelize KBF search for brew info in shared content building

michalszynkiewicz:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/documents/sharedcontent/BrewSearcher.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/documents/sharedcontent/BrewSearcher.java
@@ -20,12 +20,16 @@ package org.jboss.pnc.bacon.pig.impl.documents.sharedcontent;
 
 import com.redhat.red.build.koji.model.xmlrpc.KojiTagInfo;
 import org.jboss.pnc.bacon.pig.impl.utils.BuildFinderUtils;
+import org.jboss.pnc.bacon.pig.impl.utils.GAV;
 import org.jboss.pnc.build.finder.koji.KojiBuild;
+import org.jboss.pnc.build.finder.koji.KojiLocalArchive;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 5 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
```diff
@@ -169,10 +170,14 @@ public static BuildConfig getKojiBuildFinderConfig() {
...
-        BuildConfig config = getKojiBuildFinderConfig();
         List<String> inputs = Collections.singletonList(file.getPath());
-        ExecutorService pool = Executors.newFixedThreadPool(2);
-        DistributionAnalyzer analyzer = new DistributionAnalyzer(inputs, config);
+        return findBuilds(includeNotFound, inputs, 2);
+    }
+
+    public static List<KojiBuild> findBuilds(boolean includeNotFound, Collection<String> inputs, int poolSize) {
+        BuildConfig config = getKojiBuildFinderConfig();
+        ExecutorService pool = Executors.newFixedThreadPool(poolSize);
```
dwalluck, line 179: I'm pretty sure this isn't doing anything. Even when you set it to 50, you've still only got two tasks in there, not 50, because you're recreating the pool each time.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/documents/sharedcontent/BrewSearcher.java
```diff
@@ -37,27 +41,59 @@
...
+    public static void fillBrewData(List<SharedContentReportRow> rows) {
+        List<String> pathsForRows = rows.stream()
+                .map(row -> row.getFilePath().toAbsolutePath().toString())
+                .collect(Collectors.toList());
+
+        Map<GAV, SharedContentReportRow> rowsByGav = rows.stream()
+                .collect(Collectors.toMap(SharedContentReportRow::getGav, Function.identity()));
+
+        List<KojiBuild> builds;
+        int attempts = 3;
```
dwalluck, line 53: So, it is still failing sometimes?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
michalszynkiewicz: oh it does a lot :) the inputs collection is quite big for Quarkus and it's visible that it's indeed using 50 threads for it.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/documents/sharedcontent/BrewSearcher.java
michalszynkiewicz: I don't know, we can try to remove it, I left it from the previous version

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/BuildFinderUtils.java
dwalluck: OK, I think I didn't follow what you are doing. Are you creating a new `DistributionAnalyzer` for every file? If the input is an archive, it should only need one, but I believe it sets the number of threads in the thread pool based on the number of CPUs.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/documents/sharedcontent/BrewSearcher.java
dwalluck: I would like to know if it's fixed. Something should be done upstream to fix this, although I am not sure what at this point.

PR #439, opened 2020-09-10
[NCL-6037] Only generate nvr list when possible

thescouser89:
If the repository generation is ignored, the checksum data is never
populated. In that scenario, we shouldn't generate the nvr list during
the release phase since we don't have the required information.

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
@@ -215,13 +215,23 @@ public final class PigFacade {
                 context.getDeliverables().getNvrListName());
     }
 
+    /**
+     * Generate the nvr list from the information gathered during repository generation. If the latter was ignored, then
+     * nothing happens
+     */
     private static void generateNvrList() {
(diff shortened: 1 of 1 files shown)


rnc:
Given this caused a problem in a product build I really think we need a test/integration test here?

michalszynkiewicz:
I think that we should drop the `--skipRepo` option completely.
Most of the next steps (like e.g. license generation) don't make sense without a repository.
One can do `bacon pig configure` and then `bacon pig build` and will get the builds done.

What do you guys think? @paulgallagher75 ?

I'm merging this one but would like to discuss the above

rnc:
> I think that we should drop the `--skipRepo` option completely.
> Most of the next steps (like e.g. license generation) don't make sense without a repository.
> One can do `bacon pig configure` and then `bacon pig build` and will get the builds done.
> 
> What do you guys think? @paulgallagher75 ?
> 
> I'm merging this one but would like to discuss the above

If this has been closed/ merged perhaps that discusion should be moved to a departs ticket?

Can we try to avoid merging when there isn't tests please?

PR #440, opened 2020-09-10
Close terminal in App and fix test failures in AppTest

dwalluck:
Close terminal in App in order to restore the original state.

Fix failures in testHelp() and testProfile() due to ANSI codes in output.

Fix failure in testConfigPath() due to file separators.

Fix failure in testProfile() due to line separators.

Fix failure in testPigTempBuild() due to no exception thrown by tapSystemErr().

Use AssertJ to get meaningful assertion failure messages.

### Checklist:

### DIFF
--- a/cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
+++ b/cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
@@ -8,8 +8,10 @@ import org.junit.jupiter.api.Test;
 
 import java.io.File;
 
+import static com.github.stefanbirkner.systemlambda.SystemLambda.restoreSystemProperties;
 import static com.github.stefanbirkner.systemlambda.SystemLambda.tapSystemErr;
 import static com.github.stefanbirkner.systemlambda.SystemLambda.tapSystemOut;
+import static com.github.stefanbirkner.systemlambda.SystemLambda.tapSystemOutNormalized;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -21,9 +23,12 @@ class AppTest {
     @Test
     void testHelp() throws Exception {
         App app = new App();
-        app.run(new String[] { "-h" });
-        String text = tapSystemOut(() -> assertEquals(0, app.run(new String[] { "-h" })));
-        assertTrue(text.contains("Usage: bacon [-hovV] [-p=<configurationFileLocation>] [--profile=<profile>]"));
+        restoreSystemProperties(() -> {
+            System.setProperty("picocli.ansi", "false");
+            app.run(new String[] { "-h" });
+            String text = tapSystemOut(() -> assertEquals(0, app.run(new String[] { "-h" })));
+            assertTrue(text.contains("Usage: bacon [-hovV] [-p=<configurationFileLocation>] [--profile=<profile>]"));
+        });
     }
 
     @Test
@@ -49,11 +54,13 @@ class AppTest {
 
     @Test
     void testConfigPath() throws Exception {
+        String configPath = FilenameUtils.separatorsToSystem("/tmp/123456789");
+        String configFile = FilenameUtils.separatorsToSystem(configPath + "/config.yaml");
         String text = tapSystemErr(
                 () -> assertEquals(
                         0,
-                        new App().run(new String[] { "-p", "/tmp/123456789", "-v", "pnc", "-o", "build", "-h" })));
-        assertTrue(text.contains("Config file set from flag with profile default to /tmp/123456789/config.yaml"));
+                        new App().run(new String[] { "-p", configPath, "-v", "pnc", "-o", "build", "-h" })));
+        assertTrue(text.contains("Config file set from flag with profile default to " + configFile));
     }
 
     @Test
@@ -149,38 +156,41 @@ class AppTest {
     void testProfile() throws Exception {
         File pncClasses = new File(App.class.getClassLoader().getResource("").getFile());
         File configYaml = new File(pncClasses.getParentFile().getParentFile().getParentFile(), PNC_TEST_CLASSES);
-
         App app = new App();
-        String text = tapSystemOut(
-                () -> assertEquals(
-                        0,
-                        app.run(
-                                new String[] {
-                                        "--verbose",
-                                        "pnc",
-                                        "--profile",
-                                        "foobar",
-                                        "admin",
-                                        "maintenance-mode",
-                                        "activate",
-                                        "-h" })));
 
-        assertEquals(
-                "Usage: bacon pnc admin maintenance-mode activate [-hov]\n"
-                        + "       [-p=<configurationFileLocation>] [--profile=<profile>] <reason>\n"
-                        + "This will disable any new builds from being accepted\n"
-                        + "      <reason>              Reason\n"
-                        + "  -h, --help                display this help message\n"
-                        + "  -o                        use json for output (default to yaml)\n"
-                        + "  -p, --configPath=<configurationFileLocation>\n"
-                        + "                            Path to PNC configuration folder\n"
-                        + "      --profile=<profile>   PNC Configuration profile\n"
-                        + "  -v, --verbose             Verbose output\n" + "\n" + "Example:\n"
-                        + "$ bacon pnc admin maintenance-mode activate \"Switching to maintenance mode for\n"
-                        + "upcoming migration\"\n",
-                text);
+        restoreSystemProperties(() -> {
+            System.setProperty("picocli.ansi", "false");
+            String text = tapSystemOutNormalized(
+                    () -> assertEquals(
+                            0,
+                            app.run(
+                                    new String[] {
+                                            "--verbose",
+                                            "pnc",
+                                            "--profile",
+                                            "foobar",
+                                            "admin",
+                                            "maintenance-mode",
+                                            "activate",
+                                            "-h" })));
 
-        text = tapSystemErr(
+            assertEquals(
+                    "Usage: bacon pnc admin maintenance-mode activate [-hov]\n"
+                            + "       [-p=<configurationFileLocation>] [--profile=<profile>] <reason>\n"
+                            + "This will disable any new builds from being accepted\n"
+                            + "      <reason>              Reason\n"
+                            + "  -h, --help                display this help message\n"
+                            + "  -o                        use json for output (default to yaml)\n"
+                            + "  -p, --configPath=<configurationFileLocation>\n"
+                            + "                            Path to PNC configuration folder\n"
+                            + "      --profile=<profile>   PNC Configuration profile\n"
+                            + "  -v, --verbose             Verbose output\n" + "\n" + "Example:\n"
+                            + "$ bacon pnc admin maintenance-mode activate \"Switching to maintenance mode for\n"
+                            + "upcoming migration\"\n",
+                    text);
+        });
+
+        String text2 = tapSystemErr(
                 () -> assertEquals(
                         0,
                         app.run(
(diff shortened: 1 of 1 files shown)


cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
```diff
@@ -76,90 +85,88 @@ void testJSONDisabled() throws Exception {
...
+                                "pig",
+                                "configure",
+                                "--releaseStorageUrl",
+                                "http://www.example.com",
+                                "--tempBuild",
+                                buildConfig.toString() }));
+
+        assertThat(text).satisfiesAnyOf(
+                t -> assertThat(t).contains("Unknown option: '--tempBuild'"),
+                t -> assertThat(t).contains("Keycloak authentication failed"));
```
dwalluck, line 116: I think there is something wrong with this test. There is never any exception thrown (`FatalException` or otherwise) even though stderr contains "Keycloak authentication failed" in the exception stack trace in the output (so the `tapSystemErr()` is swallowing it?). I never see the message about unknown option, but if the command line fails to parse, how would it be getting far enough to throw an exception in the first place? What is the intended output of the test?

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -137,48 +137,46 @@ public int run(String[] args) {
...
-                    line = reader.readLine(prompt, null, (MaskingCallback) null, null);
-                    systemRegistry.execute(line);
-                } catch (UserInterruptException e) {
-                    // Ignore
-                } catch (EndOfFileException e) {
-                    AnsiConsole.systemUninstall();
-                    return 0;
-                } catch (Exception e) {
-                    systemRegistry.trace(e);
+            try (Terminal terminal = TerminalBuilder.builder().build()) {
```
rnc, line 141: This was not in the picocli example - have you added a ticket there?

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
```diff
@@ -76,90 +85,88 @@ void testJSONDisabled() throws Exception {
...
+                                            "--verbose",
+                                            "pnc",
+                                            "--profile",
+                                            "foobar",
+                                            "admin",
+                                            "maintenance-mode",
+                                            "activate",
+                                            "-h" })).isZero());
+            assertThat(text2).containsOnlyOnce("Config file set from flag with profile foobar to");
+        });
```
rnc, line 170: I am unclear as to why you are changing these tests. They worked fine in my local laptop, and on the CI tests. I am not convinced about adding an extra library and on the whole I do not see the reason for this PR.

rnc (changes requested):
On the whole its a -1 from me. Apart from the terminal change (which is a +1 if picocli issue was submitted), the rest I am not convinced by.

cli/pom.xml
```diff
@@ -42,6 +42,11 @@
             <groupId>info.picocli</groupId>
             <artifactId>picocli-shell-jline3</artifactId>
         </dependency>
+        <dependency>
```
dwalluck, line 45: This is not a new dependency. It is listed in the parent POM already. The issue is if you do `assertTrue(String.contains())` you only get a boolean `true` or `false` and no contextual information about what failed. When a test fails, you don't really know why. The AssertJ method gives you both the expected and actual result, maybe even the position where it failed (unless that's just a a Hamcrest feature).

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
```diff
@@ -1,47 +1,56 @@
...
     private static final String PNC_TEST_CLASSES = FilenameUtils.separatorsToSystem("pnc/target/test-classes");
 
     @Test
     void testHelp() throws Exception {
-        App app = new App();
-        app.run(new String[] { "-h" });
-        String text = tapSystemOut(() -> assertEquals(0, app.run(new String[] { "-h" })));
-        assertTrue(text.contains("Usage: bacon [-hovV] [-p=<configurationFileLocation>] [--profile=<profile>]"));
+        restoreSystemProperties(() -> {
+            System.setProperty("picocli.ansi", "false");
```
dwalluck, line 20: Setting `picocli.ansi=false`, otherwise it is possible that ANSI codes apepar int he output and the contains fails. This is also recommended to do by the picocli author https://github.com/remkop/picocli/pull/1166#issuecomment-690001747.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
```diff
@@ -1,47 +1,56 @@
...
+                                        "-p",
+                                        FilenameUtils.separatorsToSystem("/tmp/123456789"),
+                                        "-v",
+                                        "pnc",
+                                        "-o",
+                                        "build",
+                                        "-h" })).isZero());
+        assertThat(text).contains(
+                "Config file set from flag with profile default to "
+                        + FilenameUtils.separatorsToSystem("/tmp/123456789/config.yaml"));
```
dwalluck, line 53: All of the other paths were (correctly) converted using `FilenameUtils.separatorsToSystem()`. However, this one was missed. Hardcoding the path here is fragile in general and should probably use `@TempDir` https://junit.org/junit5/docs/5.4.0/api/org/junit/jupiter/api/io/TempDir.html.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
```diff
@@ -76,90 +85,88 @@ void testJSONDisabled() throws Exception {
...
+                            app.run(
+                                    new String[] {
+                                            "--verbose",
+                                            "pnc",
+                                            "--profile",
+                                            "foobar",
+                                            "admin",
+                                            "maintenance-mode",
+                                            "activate",
+                                            "-h" })).isZero());
```
dwalluck, line 138: Incorrect use of hardcoded line separators here instead of what is actually used, `System.lineSeparator()`. It is also possible to strip them by replacing `tapSystemOut()` with `tapSystemOutNormalized()`.

dwalluck:
Filed issue about jline3 example and terminal upstream at picocli https://github.com/remkop/picocli/issues/1168.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: I thought you were on Linux now? I think this might be helpful to discuss offline

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: I was just examining the log for that profile but ok

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: This passed on ci though....

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
dwalluck: I didn't actually check if `-p` creates the file or not. If it's not actually creating it, just logging it, then it's OK to set it to something "fake" like you have it. But, I can also change it to use `@TempDir` easily.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: Because all this passed on the ci

dwalluck: I have asked the author for clarification, but his own tests is setting this via a `@Rule` which is the JUnit 4 equivalent of this lambda.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
dwalluck: Perhaps you mean only to run your tests in one particular environment then:

```@EnabledIfSystemProperty(named = "line.separator", matches = "\\n")```

or

```
@EnabledOnOs(value = {OS.LINUX})
```

dwalluck: The `EnabledOnOs` is not so good as what determines the correctness of the test is not at all what OS it is on, but what the code under test, in this case picocli, outputs. If picocli outputs `\n`, then hardcoding `\n` in the test would be 100% correct. If picocli outputs `System.lineSeparator()`, then you want to test for `System.lineSeparator()`. The point is to match the expected output of picocli in this case.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
dwalluck: Author responded with the various cases https://github.com/remkop/picocli/pull/1166#issuecomment-690822211

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
dwalluck: My PR was merged upstream https://github.com/remkop/picocli/pull/1169

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: I think you have just set this on two of the tests - would it not be better as a BeforeAll/AfterAll ?

cli/pom.xml
rnc: I guess this is personal preference. If the tests are passing, then its just a nicety.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
rnc: Ok.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
rnc: Hmm, yes I see, its both a flaw in the test and the code. Will look into it.

rnc: Improved in https://github.com/project-ncl/bacon/pull/452

cli/pom.xml
dwalluck: An example failure as it is now is `AppTest.testHelp:26 expected: <true> but was: <false>`. But it's comparing `String`s, so in terms of seeing why it failed. Seeing the strings would help. The normal `assertEquals` does this, but doesn't work for substrings.

I believe it's in Hamcrest, not AssertJ, but it would actually show something like `expected: "foo bar" but was "foo b[ox]"`. Now, instead of true/false you see the exact reason and location it failed.

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
dwalluck: Yes, this makes sense, in the sense it should work similar to how `@Rule` works, but it doesn't look like I can put a lambda in those methods. Doesn't the code you want to run have to be inside the lambda since there's no on/off, it just resets it after the lambda completes?

cli/src/test/java/org/jboss/pnc/bacon/cli/AppTest.java
dwalluck: That will do `return tapSystemOut(statement).replace(lineSeparator(), "\n");` This is a minimal one-word change then instead of changing every line in that test.

rnc: That sounds better then....

dwalluck:
I am going to reforce this PR with _much_ less changes.

I still do think it's impossible to tell why the tests failed when using `assertTrue()`, but it seems I may prefer Hamcrest to AssertJ, anyway. But, now that I can see why they are failing, I don't really need to keep it.

dwalluck:
@rnc Completely redone with minimal changes. I will open another PR for `App.java`.

rnc (approved):

PR #442, opened 2020-09-10
retry downloads

michalszynkiewicz:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
@@ -29,6 +29,7 @@ import org.jboss.pnc.bacon.pig.impl.out.PigReleaseOutput;
 import org.jboss.pnc.bacon.pig.impl.out.PigRunOutput;
 import org.jboss.pnc.bacon.pig.impl.pnc.ImportResult;
 import org.jboss.pnc.bacon.pig.impl.repo.RepositoryData;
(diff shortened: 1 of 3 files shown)


rnc:
Can FileDownLoadUtils ever be called by multiple threads? i.e. could the `attempts` value be corrupted? Otherwise looks good to me.

thescouser89:
Haa just a silly suggestion, but should we set the default attempts to a value greater than 1? Just to have some fault tolerance on by default.

Also tiny tiny tiny bug: if i set the attempts to be negative that could potentially create an infinite loop if the stuff we are downloading keeps failing.

dwalluck:
> Haa just a silly suggestion, but should we set the default attempts to a value greater than 1? Just to have some fault tolerance on by default.

Why not?

> Also tiny tiny tiny bug: if i set the attempts to be negative that could potentially create an infinite loop if the stuff we are downloading keeps failing.

OK so `if retries <= 0: retries = 1 (or default_retries)`

michalszynkiewicz:
> > Haa just a silly suggestion, but should we set the default attempts to a value greater than 1? Just to have some fault tolerance on by default.
> 
> Why not?

I'll set it to `2` for pig.

> 
> > Also tiny tiny tiny bug: if i set the attempts to be negative that could potentially create an infinite loop if the stuff we are downloading keeps failing.
> 
> OK so `if retries <= 0: retries = 1 (or default_retries)`

Good catch!

thescouser89:
:+1:

PR #444, opened 2020-09-10
pig test against real PNC

michalszynkiewicz:
### Checklist:

* [X] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) for your change if user-facing?
* [X] Have you added -unit- tests for your change?

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -16,6 +16,7 @@
         <keycloak-user>user</keycloak-user>
         <pnc-url>http://localhost:8080</pnc-url>
         <scm-host>example.com</scm-host>
+        <testTags>!RealServiceOnly</testTags>
     </properties>
     <dependencies>
         <!-- Project deps -->
@@ -93,7 +94,27 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <groups>${testTags}</groups>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>functionalTests</id>
+            <activation>
+                <property>
+                    <name>functionalTests</name>
+                </property>
+            </activation>
+            <properties>
+                <testTags>!MockOnly</testTags>
+            </properties>
+        </profile>
+    </profiles>
 
 </project>
--- a/integration-tests/src/test/java/org/jboss/pnc/bacon/test/TestType.java
+++ b/integration-tests/src/test/java/org/jboss/pnc/bacon/test/TestType.java
@@ -1,0 +1,15 @@
+package org.jboss.pnc.bacon.test;
+
+/**
+ * If a test is supposed to be executed either only with mocks or only with real service, add one of the tags to the
+ * test
+ *
+ * If the test should run regardless of environment, no tag is required
+ */
+public class TestType {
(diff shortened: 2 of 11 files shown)


michalszynkiewicz:
the change also adds a posibility to overwrite config varialbes with system properties

michalszynkiewicz:
It's much easier to write such tests against real PNC, and it also should make it much easier to reproduce the bugs, like the one we have for wrong build returned when the builds are skipped.

Next we could add mocks to be able to run it without the real systems (such duality, as I understand was @janinko's original idea)

integration-tests/pom.xml
```diff
@@ -93,7 +93,39 @@
...
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <excludedGroups>RealServiceOnly</excludedGroups>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>mock-tests</id>
+                    </execution>
+                </executions>
```
rnc, line 105: Do we need another execution here?

integration-tests/pom.xml
```diff
@@ -93,7 +93,39 @@
...
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludedGroups>MockOnly</excludedGroups>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>real-tests</id>
+                            </execution>
```
rnc, line 123: Won't this run all the tests again?Don't we just want to run the functional test ones in this execution?

rnc:
> It's much easier to write such tests against real PNC, and it also should make it much easier to reproduce the bugs, like the one we have for wrong build returned when the builds are skipped.
> 
> Next we could add mocks to be able to run it without the real systems (such duality, as I understand was @janinko's original idea)

Yes I agree, I often find most of the time is spent writing a test to reproduce the issue :-D 

Looks good, couple of comments about how tests are selected via maven.

dwalluck:
We _could_ switch to using surefire for unit tests and failsafe for integration tests, then it's easier to just specify

```
<configuration>
    <groups>SomeGroup</groups>
</configuration>
```

one type of group (UnitTest, IntegrationTest, for example) in each plugin and no executions should be necessary.

integration-tests/pom.xml
michalszynkiewicz: true, I haven't thought it through.

Looks like there needs to be another profile for mock only and all other tests could be executed in the "normal" execution. I'll update the PR

rnc: I would view all other tests, including mock as the default. Perhaps just run the real tests in the separate profile ?

michalszynkiewicz:
I found a way of configuring surefire in a way that works ;)

The tests I marked as MockOnly are failing when tested with stage, if that's fixed it's enough to remove the tag from them to make them executed in both cases.

integration-tests/pom.xml
michalszynkiewicz: no new execution now :)

integration-tests/pom.xml
michalszynkiewicz: Now by default all tests by the real service tests are executed, and if switched to `functionalTests`, only the tests that work well with mocks are excluded.

michalszynkiewicz:
I'm merging it, it doesn't impact "normal usage". Happy to refactor later if required

PR #447, opened 2020-09-10
Add retries when downloading links on bacon_install.py

thescouser89:
When downloading artifacts from repositories.jboss.org, it sometimes
fails because the specific server we are hitting on
repositories.jboss.org does not have the content yet, and returns 404.
The retries is an attempt to workaround this by retrying to download the
artifact up to 7 times until we give up.

### Checklist:

### DIFF
--- a/bacon_install.py
+++ b/bacon_install.py
@@ -121,9 +121,12 @@ def download_maven_metadata_xml(url, folder):
     download_link(link, folder, "maven-metadata.xml")
 
 
-def download_link(link, folder, filename):
+def download_link(link, folder, filename, retries=7):
     """
     Download the link into the folder with name 'filename'
+
+    Add retries in case the download fails due to stale server content issue on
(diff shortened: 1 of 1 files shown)


dwalluck:
Did you want to sleep between retires or just rapid fire them? There is this https://pypi.org/project/backoff/ but I don't think you want to add a dependency!

thescouser89:
Yeah sleep is not a bad idea! We could just run a simple exponential sleep based on how many retries? WDYT?

dwalluck:
Sure, however you want to do it. I think even constant would help. It's playing nice with the server and I also think if you get a failure and all of the retires are within a couple milliseconds of the failure then they are probably more likely to fail, too.

thescouser89:
This PR should incorporate your great idea: https://github.com/project-ncl/bacon/pull/454

PR #449, opened 2020-09-10
NCL-6033: List operations on bacon are broken

dwalluck:
### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractBuildListCommand.java
+++ b/common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractBuildListCommand.java
@@ -17,11 +17,12 @@
  */
 package org.jboss.pnc.bacon.common.cli;
 
-import org.jboss.pnc.client.RemoteCollection;
 import org.jboss.pnc.client.RemoteResourceException;
 import org.jboss.pnc.dto.Build;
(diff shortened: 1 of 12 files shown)


dwalluck:
I think this is what was decided on. It's hard to add tests with no live test structure in place prior to this PR.

It;s also possible to do offline tests using wiremock. You have to set it up to capture some output of the commands first.

rnc (approved):
Looks good! I suspect there is no obvious way of testing this (without the work that @michalszynkiewicz  is putting in #444 )

dwalluck:
I know how to get the wiremock captures, but it takes time. They have the advantage of being offline (except localhost), but they're harder to write.

If it requires a live server, then it's only going to be able to run in Jenkins or with additional config and access to internal servers, but they're easier to write as there is no mocking beforehand.

rnc:
> I know how to get the wiremock captures, but it takes time. They have the advantage of being offline (except localhost), but they're harder to write.
> 

*Much* harder! If manual testing is ok, then I would vote to merge.

dwalluck:
@thescouser89 didn't manually test last time which is how we didn't find out until later. He may want to try first this time.

thescouser89 (approved):
LGTM! Compiled locally and the list operations are working again!!! :)

PR #450, opened 2020-09-10
download retry fixes

michalszynkiewicz:
### Checklist:

* [X] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) for your change if user-facing?

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileDownloadUtils.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileDownloadUtils.java
@@ -36,8 +36,6 @@ import java.net.URI;
  *         Date: 4/16/18
  */
 public class FileDownloadUtils {
-    private FileDownloadUtils() {
-    }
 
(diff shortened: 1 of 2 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileDownloadUtils.java
```diff
@@ -99,6 +99,14 @@ public static void setAttempts(int attempts) {
             attempts = MAX_ATTEMPTS;
         }
 
+        if (attempts <= 0) {
+            log.warn("Number of download attempts has to be a positive integer. Setting to {}", 1);
```
dwalluck, line 103: The log message should use `, DEFAULT_ATTEMPTS` not `1`, in case you decide change it later.

michalszynkiewicz: :+1:

thescouser89:
Thanks for the changes!!! :)

PR #453, opened 2020-09-11
Close Terminal in App

dwalluck:
This came from an upstream picocli example, which I have fixed upstream, see https://github.com/remkop/picocli/issues/1168 and https://github.com/remkop/picocli/pull/1169.


rnc:
LGTM

PR #455, opened 2020-09-11
Use assertj for testing strings for better failure messages

dwalluck:
It's OK if you still prefer the other way, but I thought I'd try a more minimal PR for AssertJ.

The advantage is more meaningful messages on test failures. For `assertEquals()` there is really no advantage to changing that, so I left `assertEquals()` and replaced:

- `assertTrue(text.contains())` with `assertThat(text).contains()`
- `assertFalse(text.contains())` with `assertThat(text).doesNotContain()`
- `assertEquals(StringUtils.countMatches(text, str), 1)` with `assertThat(text).containsOnlyOnce(str)`

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -43,6 +43,11 @@
             <artifactId>picocli-shell-jline3</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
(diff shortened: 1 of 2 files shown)


rnc:
I didn't mean I prefer one way or the other, just that it didn't seem the effort of changing them. But if you've already done the PR here, thats fine by me.

PR #458, opened 2020-09-12
[NCL-6044] Check if all builds in final state when group build done 

thescouser89:
Due to internal logic in PNC-Orch, a group build gets updated in the
database just before the final build in the group build is done. This
may cause the scenario where during a time window, the group build state
is successful, but there's still 1 build in the group build still in
running state.

The situation is even worse if the group build ends in state
'NO_REBUILD_REQUIRED'. Before the builds even have a chance to be
scheduled, the group build is already marked as 'NO_REBUILD_REQUIRED'
beforehand. Then when we get the list of builds, they may still be in
status RUNNING. This is dangerous because then the final artifact list
may be empty, and in the case of NO_REBUILD_REQUIRED, we'll never get
the original successul build.

To avoid these scenarios, we also check if all the builds in a group
build are in a final state once the group build is marked as successful.

### Checklist:

### DIFF
--- a/pig/pom.xml
+++ b/pig/pom.xml
@@ -139,6 +139,11 @@
             <artifactId>mockito-inline</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jeasy</groupId>
+            <artifactId>easy-random-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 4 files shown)


michalszynkiewicz (approved):

thescouser89:
I don't think I can provide any tests for that :(

thescouser89:
I've moved it back to draft to see if I can add some tests :) Will do on Tuesday. Sorry for the delay on getting it ready.

PR #459, opened 2020-09-12
NCL-6053: pig - .bacon dir in a temp loc

michalszynkiewicz:
### Checklist:

* [X] Have you added unit tests for your change?

### DIFF
--- a/integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
+++ b/integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
@@ -14,6 +14,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.file.Path;
 import java.util.Base64;
 import java.util.Random;
 
@@ -64,28 +65,33 @@ public class AbstractTest {
     }
 
     protected void execute(String... args) {
-        ExecutionResult result = executor.runCommand(args);
+        ExecutionResult result = executor.runCommand(null, args);
         assertThat(result.getOutput()).isEmpty();
         assertThat(result.getError()).isEmpty();
         assertThat(result.getRetval()).isZero();
     }
 
     protected <T> T executeAndDeserialize(Class<T> clazz, String... args) throws JsonProcessingException {
-        ExecutionResult result = executor.runCommand(args);
+        ExecutionResult result = executor.runCommand(null, args);
         log.debug("stderr:{}{}", System.lineSeparator(), result.getError());
         assertThat(result.getRetval()).isZero();
         return result.fromYAML(clazz);
     }
 
-    protected <T> T executeAndDeserializeJSON(Class<T> clazz, String... args) throws JsonProcessingException {
-        ExecutionResult result = executor.runCommand(args);
+    protected <T> T executeAndDeserializeJSON(Class<T> clazz, Path workingDir, String... args)
+            throws JsonProcessingException {
+        ExecutionResult result = executor.runCommand(workingDir, args);
         log.debug("stderr:{}{}", System.lineSeparator(), result.getError());
         assertThat(result.getRetval()).isZero();
         return result.fromYAML(clazz);
     }
 
     protected ExecutionResult executeAndGetResult(String... args) {
-        return executor.runCommand(args);
+        return executor.runCommand(null, args);
+    }
+
+    protected ExecutionResult executeAndGetResult(Path workingDir, String... args) {
+        return executor.runCommand(workingDir, args);
     }
 
 }
--- a/integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/PigFunctionalTest.java
+++ b/integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/PigFunctionalTest.java
@@ -2,7 +2,6 @@ package org.jboss.pnc.bacon.test.pig;
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.jboss.pnc.bacon.common.Constant;
-import org.jboss.pnc.bacon.common.exception.FatalException;
 import org.jboss.pnc.bacon.config.Config;
 import org.jboss.pnc.bacon.config.PigConfig;
 import org.jboss.pnc.bacon.config.Validate;
@@ -14,6 +13,8 @@ import org.junit.jupiter.api.Tag;
 
 import java.nio.file.Path;
 import java.util.Optional;
+
+import static org.jboss.pnc.bacon.common.Constant.PIG_CONTEXT_DIR;
 
 @Tag(TestType.REAL_SERVICE_ONLY)
 public abstract class PigFunctionalTest {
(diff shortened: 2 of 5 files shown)


integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
```diff
@@ -64,28 +65,33 @@ protected static String getRandomString() {
...
+            throws JsonProcessingException {
+        ExecutionResult result = executor.runCommand(workingDir, args);
         log.debug("stderr:{}{}", System.lineSeparator(), result.getError());
         assertThat(result.getRetval()).isZero();
         return result.fromYAML(clazz);
     }
 
     protected ExecutionResult executeAndGetResult(String... args) {
-        return executor.runCommand(args);
+        return executor.runCommand(null, args);
```
dwalluck, line 90: So this sets the `dir` argument of `Runtime.exec()` to `null` which means `$CWD`, but I still don't really like that it will create this dir anywhere I run it. I think picking a single location would be better.

The other issue is because it's not a unique location, it would be possible to run the `run` with one project and `release` with another project using the same `.bacon/` directory, and what would happen then?

I kind of like `$HOME/.bacon/<project_id>/` or something similar.

rnc:
What is the bacon dir used for? Is it needed by users after a run?

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
rnc: Great points - I think we need to understand its purpose before choosing a solution.

dwalluck:
> What is the bacon dir used for? Is it needed by users after a run?

I believe it stores the state for a particular `run` and is then used when you do a `release`.

But, that makes it specific to a particular `run`, and you won't know without examining the file which `run` this belongs to. It is also a hidden directory and created in whatever directory you happen to be in when you run it. Obviously, if it's a read-only directory for your user it's just going to fail. It doesn't create it where it's known to be writable.

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
michalszynkiewicz: By default the .bacon directory it's created where the program is executed. The location can be overriden with an environment varialbe and a system property.
Current directory was chosen as a default to allow running the tool for multiple projects separately.

If the config yaml or any other file in the config dir would change, the context will be dropped (the previous one won't be read and will be overwritten at the end of the execution). It's checked by storing a hash sum of all files in the directory (and their locations, moving a file is also considered a change).

`executeAndGetResult` is also used for non-pig tests, for it I haven't changed the working directory (hence the null here)

michalszynkiewicz:
pig context stores data between executions of pig.
With it one can do `pig configure` and then do `pig build` and data from one would be passed to another. As written in the review comment, it's invalidated if configuration has been changed between the executions.

dwalluck:
What data specifically can be shared between runs? In any case, a change in project should cause the configuration to be invalidated. It seems easier to me to just store the configuration under some unique key derived from the project name or something else.

I noticed also having to set the pnc configuration file via an environment variable. This is a bit odd and not usually how other programs work. I'd expect to set it via a program flag, if anything, as the defaults should also be reasonable to begin with.

rnc:
@michalszynkiewicz Interesting, thanks. How does it know whether to invalidate it or not? I wonder if it would be better to have this as an option flag - it could default to , as @dwalluck suggested $Home/.bacon but can be overridden to use a current directory. This would solve the problem with these hidden directories being left lying around, would still allow the same flexibility and the tests could use it with a temp dir.

Also... is this documented?!

michalszynkiewicz:
there's a `--clean` option that invalidates the context too.
I like the current working directory better but we can create a JIRA and discuss it. It is not really relevant to this PR ;)

Re using env variable for the location, I think it's better than system property or parameter because you can set it in your `.bashrc` or `.profile` and don't have to pass it every time. 
It's also not an uncommon thing in the java world, e.g. spring profile can be selected this way, configuration of apps running in kubernetes pods is also often passed this way.

michalszynkiewicz:
Regarding how does it know, I calculate a hash of the pig configuration directory.

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
rnc: I think it would be better to use a Junit tmp dir here and pass that through?

integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/PigFunctionalTest.java
```diff
@@ -29,13 +30,15 @@ static String init(Path configDir, boolean clean, Optional<String> releaseStorag
         String suffix = prepareSuffix();
         // todo release storage url mocking
         if (configDir == null) {
-            throw new FatalException("You need to specify the configuration directory!");
+            throw new RuntimeException("You need to specify the configuration directory!");
```
rnc, line 33: FatalException extends runtimeexception - why the change here?

integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/PigFunctionalTest.java
```diff
@@ -29,13 +30,15 @@ static String init(Path configDir, boolean clean, Optional<String> releaseStorag
...
         }
+        System.setProperty(PIG_CONTEXT_DIR, targetDirectory.toString());
```
rnc, line 35: I think tests that set system properties should use https://github.com/stefanbirkner/system-lambda (https://github.com/stefanbirkner/system-rules for JUnit4)

integration-tests/src/test/java/org/jboss/pnc/bacon/test/AbstractTest.java
rnc: Ah, I've just seen down below you are passing a TempDir in :-) What about changing for all tests so they consistently run from a tempDirectory - or do some require a certain context pwd?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/PigContext.java
```diff
@@ -196,7 +196,7 @@ private static PigContext readContext(boolean clean, Path configDir) {
         String sha = hashDirectory(configDir);
 
         PigContext result;
-        String ctxLocationEnv = System.getenv(PIG_CONTEXT_DIR);
+        String ctxLocationEnv = System.getProperty(PIG_CONTEXT_DIR, System.getenv(PIG_CONTEXT_DIR));
```
rnc, line 199: For reference, for perhaps a future consideration, picocli supports default values of e.g. environment variables ( https://picocli.info/#_variable_interpolation ) so it could be a parameter as well.

rnc:
> there's a `--clean` option that invalidates the context too.
> I like the current working directory better but we can create a JIRA and discuss it. It is not really relevant to this PR ;)
> 
> Re using env variable for the location, I think it's better than system property or parameter because you can set it in your `.bashrc` or `.profile` and don't have to pass it every time.
> It's also not an uncommon thing in the java world, e.g. spring profile can be selected this way, configuration of apps running in kubernetes pods is also often passed this way.

Could you create a separate JIRA then please? As mention in one of my comments, picocli supports parameters with environment variables I think - that also could be a separate ticket perhaps?

PR #466, opened 2020-09-16
[NCL-6043] noWait support for bacon pig build

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
@@ -89,7 +89,7 @@ public final class PigFacade {
         return pncImporter.readCurrentPncEntities();
     }
 
-    public static GroupBuildInfo build(boolean tempBuild, boolean tempBuildTS, RebuildMode rebuildMode) {
+    public static GroupBuildInfo build(boolean tempBuild, boolean tempBuildTS, RebuildMode rebuildMode, boolean wait) {
         ImportResult importResult = context().getPncImportResult();
         if (importResult == null) {
             importResult = readPncEntities();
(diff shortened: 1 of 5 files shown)


rnc:
Should we not also have a test where we _do_ wait?

jomrazek:
Do you mean opposite? all tests should be like before so "waiting" for results

rnc:
> Do you mean opposite? all tests should be like before so "waiting" for results

No, on top of. So while all tests are as before, surely with new functionality we should look at adding new tests for it?

jomrazek:
I would understand that, but in your first comment you mentioned that we should have tests where we **do** wait, we're waiting in all of build tests. I could create test where we **don't** wait although I don't see much value in it as we return just message that build was started.

rnc:
> I would understand that, but in your first comment you mentioned that we should have tests where we **do** wait, we're waiting in all of build tests. I could create test where we **don't** wait although I don't see much value in it as we return just message that build was started.

Ok apologies, I meant the inverse :-)

pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
```diff
@@ -100,8 +104,12 @@ public static GroupBuildInfo build(boolean tempBuild, boolean tempBuildTS, Rebui
         }
 
         GroupBuild groupBuild = new PncBuilder()
-                .buildAndWait(importResult.getBuildGroup(), tempBuild, tempBuildTS, rebuildMode);
-        return new BuildInfoCollector().getBuildsFromGroupBuild(groupBuild);
+                .buildAndMaybeWait(importResult.getBuildGroup(), tempBuild, tempBuildTS, rebuildMode, noWait);
```
michalszynkiewicz, line 107: please rename it to just `build` :)

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
```diff
@@ -60,13 +60,16 @@ public PncBuilder() {
...
+    public GroupBuild buildAndMaybeWait(
             GroupConfigurationRef group,
             boolean tempBuild,
             boolean tempBuildTS,
-            RebuildMode rebuildMode) {
+            RebuildMode rebuildMode,
+            boolean noWait) {
         GroupBuild groupBuild = run(group, tempBuild, tempBuildTS, rebuildMode);
-        waitForSuccessfulFinish(groupBuild.getId());
+        if (!noWait) {
```
michalszynkiewicz, line 70: just a note, I think it would be a bit clearer if the method param was the inverse, `wait`, and not `noWait`

jomrazek:
@michalszynkiewicz thanks for feedback, PR updated

michalszynkiewicz:
:+1:

PR #467, opened 2020-09-16
Enable history in bacon shell

rnc:
### Checklist:

### DIFF
--- a/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
+++ b/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
@@ -31,7 +31,7 @@ import org.jboss.pnc.bacon.pnc.Pnc;
 import org.jline.builtins.Builtins;
 import org.jline.builtins.SystemRegistry;
 import org.jline.builtins.SystemRegistryImpl;
-import org.jline.builtins.Widgets;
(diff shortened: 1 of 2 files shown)


rnc:
@michalszynkiewicz Added your shell history for you. Couldn't see a way of enabling that extra prompt by default so added a log line for now.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -143,17 +143,16 @@ public int run(String[] args) {
                         .terminal(terminal)
                         .completer(systemRegistry.completer())
                         .parser(parser)
+                        .variable(LineReader.HISTORY_FILE, Constant.HISTORY)
+                        .variable(LineReader.HISTORY_FILE_SIZE, 50)
```
michalszynkiewicz, line 147: is that last 50 entries in the history file? I'd like much more :)

michalszynkiewicz (approved):

michalszynkiewicz:
Thanks a lot @rnc !

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
rnc: Er, sure, sorry just picked a random number :-)

PR #469, opened 2020-09-17
Enable tailtip widget by default by explicitly overriding picicli jline3

rnc:
This explicitly enables tailtip support for the console interface. To do that I have had to inline a lot of picocli-jline3 support until https://github.com/remkop/picocli/issues/1175 is fixed (found during discussions on https://github.com/jline/jline3/issues/571 ).

If this is ok , I'll add a separate internal tracker to remove the override code (the second commit) once picocli 4.6 is out.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -43,6 +43,14 @@
             <artifactId>picocli-shell-jline3</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jline</groupId>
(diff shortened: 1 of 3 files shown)


thescouser89:
To be honest, I have no idea this tailtip thing existed and what the code does. But I did checkout the commit and built it and it looks really cool.

Feel free to create an internal tracker to remove the override code when picocli 4.6 is out. And thanks for the commit! :)

PR #470, opened 2020-09-17
[NCL-6075] Bacon jsonOutput flag went missing

jomrazek:
### Checklist:

### DIFF
--- a/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
+++ b/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
@@ -74,7 +74,10 @@ public class App {
     }
 
     @SuppressWarnings("FieldMayBeFinal")
-    @Option(names = Constant.JSON_OUTPUT, description = "use json for output (default to yaml)", scope = INHERIT)
+    @Option(
+            names = { Constant.JSON_OUTPUT, "--jsonOutput" },
+            description = "use json for output (default to yaml)",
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/jboss/pnc/bacon/common/cli/JSONCommandHandler.java
```diff
@@ -19,9 +19,14 @@
...
-        if (option == null) {
+        CommandLine.Model.OptionSpec optionFull = commandSpec.root().findOption(Constant.JSON_FULL_OUTPUT);
+        if (option == null && optionFull == null) {
             throw new FatalException("Unable to locate JSON parameter in root");
         }
+        if (option == null) {
+            log.debug("JSON command is enabled: " + optionFull.getValue());
+            return optionFull.getValue();
+        }
         log.debug("JSON command is enabled: " + option.getValue());
```
rnc, line 30: I don't think this is needed. We only need to find the option by name and check if it is enabled.

rnc (changes requested):

common/src/main/java/org/jboss/pnc/bacon/common/cli/JSONCommandHandler.java
jomrazek: and you're right :+1:

rnc (approved):
LGTM!

PR #472, opened 2020-09-18
[NCL-6060] Get the build log on demand

thescouser89:
Since the build log is not stored in the context.json due to size, that
causes some issues.

1. If a pig command depends on the build log and it is not present, this
   will stop the command from working. (e.g for `pig addons`)

It's also better to only get the build logs if necessary since they are
huge and take a lot of space. The build logs are cached within the same
pig command, but if you run multiple pig commands, the build logs will
be loaded from the PNC server again.

### Checklist:


thescouser89:
Testing a static class is hard :(

PR #482, opened 2020-09-23
[NCL-5982] Add validation to build-config.yaml

thescouser89:
This uses Hibernate Validator and Java EE Bean validation to accomplish
this.

Two custom bean validators are written to validate Build Configs and the
flow generation when set to 'DOWNLOAD'.

e.g
```
~/projects/work/bacon ncl-5982 ❯ java -jar cli/target/bacon.jar pig build nightly                                                           8s
[INFO ] - the configuration has been changed since the last run, using clean pig context
[INFO ] - HV000001: Hibernate Validator 2.1.2-SNAPSHOT
[ERROR] - Errors while validating the build-config.yaml:
Field: builds[0].project must not be blank
Field: product.name must not be blank
```

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
@@ -32,6 +32,8 @@ import org.jboss.pnc.enums.BuildType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.validation.constraints.NotBlank;
+
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Field;
 import java.net.URLDecoder;
@@ -65,12 +67,12 @@ public class BuildConfig {
     private static final Logger log = LoggerFactory.getLogger(BuildConfig.class);
     private static final ClientCreator<EnvironmentClient> CREATOR = new ClientCreator<>(EnvironmentClient::new);
 
-    private String name;
-    private String project;
-    private String buildScript;
+    private @NotBlank String name;
+    private @NotBlank String project;
+    private @NotBlank String buildScript;
     private String scmUrl;
     private String externalScmUrl;
-    private String scmRevision;
+    private @NotBlank String scmRevision;
     private String description;
     private String environmentId;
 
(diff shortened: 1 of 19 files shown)


thescouser89:
retest this please

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
```diff
@@ -271,8 +269,9 @@ public String getEnvironmentId() {
             } catch (RemoteResourceException e) {
                 throw new FatalException("Exception while talking to PNC", e);
             }
-        } else {
-            throw new FatalException("No environmentId / environmentSystemImageId defined for the build config");
         }
+
+        // if we're here, no environmentId could be found
+        return null;
```
dwalluck, line 275: Are you returning `null` from a method that returns `boolean`? Do you mean to return `false`? For fields, if you're using `Boolean` instead of `boolean`, you should probably validate the field with `@NotNull`.

thescouser89: Oh shoot good catch. I didn't notice that I added this in the wrong area. THanks for pointing this out.

thescouser89: @dwalluck I relooked at the code and it's Github that is confusing you into thinking this is for method `isUpgradableFrom`. It's for the `getEnvironmentId` that I'm returning null.

dwalluck: Oops! I even double-checked and the line numbers seemed to match up. But I guess decide if you want to return `null` or throw an exception, in any case. I know old Java APIs like `File` used to return `null` and then they supposedly stopped doing that?

thescouser89: Yeah. I wanted to throw NPE but I think that:
1. This won't actually happen due to the validation beforehand
2. I'll probably have to refactor the `getEnvironmentId` method because it does a bit too much, doesn't it?

I'm still on the fence on what to do with this. Any advice welcome! :)

rnc:
Nice! Am I right in thinking that the annotations allow messages to be added for the validation? (e.g. to the buildscript annotation as I believe there was a case where someone hadn't filled that in). Would having a set of messages make it easier for a generic validation?

thescouser89:
Yeah the annotations do allow for custom messages to be added. I actually haven't added support for this on the custom validators, but yeah possible.

What custom message do you have in mind? I actually didn't mind the default message for missing stuff ("must not be blank")

PR #490, opened 2020-10-01
Use `executeBrewPush` only once

thescouser89:
Since `executeBrewPush` invocation disconnects the websocket client, we
shouldn't re-use the AdvancedBuildClient for further brew pushes since
the websocker listener is disconnected.

This commit instead creates a new client for every push to workaround
the limitation.

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/PigFacade.java
@@ -244,25 +244,20 @@ public final class PigFacade {
                     "Pushing the following builds to brew: {}",
                     buildsToPush.stream().map(PncBuild::getId).collect(Collectors.toList()));
         }
-        try (AdvancedBuildClient pushingClient = new AdvancedBuildClient(PncClientHelper.getPncConfiguration())) {
-            for (PncBuild build : buildsToPush) {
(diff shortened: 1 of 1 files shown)


paulgallagher75:
Tested and worked pushing vertx builds to brew

thescouser89:
CC: @michalovjan `executeBrewPush` in the `AdvancedBuildClient` is single use right now. I think we should only disconnect on the `close` method, rather than at the end of the method.

michalovjan:
@thescouser89 Thx for finding where the problem is. I think we should have a meeting about the clients, because of the closing/disconnecting causing issues everywhere.

rnc:
@michalovjan @thescouser89 Is this also related to #493 ?

michalovjan:
@rnc Sort of, the clients mentioned there are RestEasy http-clients from regular part of rest-client library which were not being closed. This is specifically an issue with websocket clients(which are also part of the rest-client library but have Advanced* prefix) which have to be closed too. I didn't want to do it through close() methods but automatically in the background. Due to the random issues I am leaning more to the close() method.

thescouser89:
@michalovjan Let's talk about it on the monday meeting! And it can spill over to the PiG meeting in 2 weeks if necessary.

PR #497, opened 2020-10-02
Fix string concatenation in log statement in JSONCommandHandler

dwalluck:
### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/bacon/common/cli/JSONCommandHandler.java
+++ b/common/src/main/java/org/jboss/pnc/bacon/common/cli/JSONCommandHandler.java
@@ -22,7 +22,8 @@ public class JSONCommandHandler {
         if (option == null) {
             throw new FatalException("Unable to locate JSON parameter in root");
         }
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/jboss/pnc/bacon/common/cli/JSONCommandHandler.java
```diff
@@ -22,7 +22,8 @@ protected boolean getJsonOutput() {
         if (option == null) {
             throw new FatalException("Unable to locate JSON parameter in root");
         }
-        log.debug("JSON command is enabled: " + option.getValue());
-        return option.getValue();
+        Boolean value = option.<Boolean> getValue();
```
thescouser89, line 25: Why the additional line here?

dwalluck: First, it should probably make one method call instead of two. Second, it could not determine the signature for `log.debug()` which is why the `<Boolean>` had to be there. Without the assignment to tell it the type, it doesn't work.

PR #498, opened 2020-10-02
Throw exception instead of returning null in refreshToken()

dwalluck:
### Checklist:


auth/src/main/java/org/jboss/pnc/bacon/auth/DirectKeycloakClientImpl.java
```diff
@@ -163,30 +165,23 @@ public Credential getCredentialServiceAccount(
...
-            return credential.toBuilder()
-                    .accessToken(response.getAccessToken())
-                    .accessTokenExpiresIn(now.plusSeconds(response.getExpiresIn()))
-                    .refreshToken(response.getRefreshToken())
-                    .refreshTokenExpiresIn(now.plusSeconds(response.getRefreshExpiresIn()))
-                    .build();
-
-        } catch (Exception e) {
-            log.error(e.getMessage());
-            return null;
```
dwalluck, line 188: Remove this and throw `KeycloakClientException` instead, because `null` was never checked for, resulting in a possible NullPointerException`, but the `KeycloakClientException` is already caught.

PR #501, opened 2020-10-07
[NCL-5980] Check for >0 build configs and builds when generating mave…

thescouser89:
…n repo

### Checklist:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepoManager.java
```diff
@@ -177,6 +177,9 @@ private RepositoryData packAllBuiltAndDependencies() {
 
     private RepositoryData buildGroup() {
         log.info("Generating maven repo for build group [{}]", pigConfiguration.getGroup());
+        if (builds.isEmpty()) {
```
dwalluck, line 180: Is it possible for either of these to be empty under normal circumstances so that you would just want to skip them and not fail?

thescouser89: Yes in weird circumstances that Paul Gallagher hit it might be empty.

PR #510, opened 2020-10-08
NCL5978 Revert changes for new PME release

rnc:
### Checklist:


dwalluck:
@rnc Do you want to unify these two versions with a property? They shouldn't ever be different right? It's creating two PRs for the same software.

rnc:
> @rnc Do you want to unify these two versions with a property? They shouldn't ever be different right? It's creating two PRs for the same software.

Er, I did?

dwalluck:
> Er, I did?

Indeed.

PR #512, opened 2020-10-08
Handle failure to create cache file directories by throwing exception

dwalluck:
The original version called `new File(configFolder).mkdirs()` without checking for failure. Then it could fail unexpectedly later. I replaced with `Files.createDirectories(path)` which throws an `IOException`. I also removed `fileExists()` and replaced with separate calls to `Files.isRegularFile()` and `Files.isDirectory()` so that we check the actual type as opposed to just the existence.

### DIFF
--- a/auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
+++ b/auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
@@ -39,29 +39,29 @@ public class CacheFile {
             String realm,
             String username,
             Credential credential) {
-
-        createConfigFolderIfAbsent();
-
         File file = new File(getCacheFile());
 
         log.debug("Writing credential to cache file {}", file);
 
         try {
+            createConfigFolderIfAbsent();
+
             Map<String, Credential> data = Collections
                     .singletonMap(generateUsernameMd5(keycloakUrl, realm, username), credential);
             CacheFile cacheFile = new CacheFile();
             cacheFile.setCachedData(data);
             mapper.writeValue(file, cacheFile);
-            setOwnerFilePermissions(getCacheFile());
+            setOwnerFilePermissions(file.toPath());
         } catch (IOException e) {
             log.error("Error saving credential to file {}", file, e);
         }
     }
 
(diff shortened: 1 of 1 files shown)


rnc:
@thescouser89 Can this be either merged or reviewed please?

auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -80,11 +81,13 @@ public static void writeCredentialToCacheFile(
...
     }
 
-    private static void createConfigFolderIfAbsent() {
-        if (!fileExists(Config.getConfigLocation())) {
+    private static void createConfigFolderIfAbsent() throws IOException {
+        Path path = Paths.get(Config.getConfigLocation());
+
+        if (!Files.isDirectory(path)) {
             String configFolder = Config.getConfigLocation();
             log.debug("Creating config folder {}", configFolder);
```
thescouser89, line 89: LGTM! Just a tiny suggestion: I guess we could just print `path.toString()` here and erase line 88? WDYT?

dwalluck: Yeah, it's basically getting `Config.getConfigLocation()` twice, once as a `Path` and then again as a `String`.

auth/src/main/java/org/jboss/pnc/bacon/auth/model/CacheFile.java
```diff
@@ -39,29 +39,29 @@ public static void writeCredentialToCacheFile(
...
         try {
+            createConfigFolderIfAbsent();
+
             Map<String, Credential> data = Collections
                     .singletonMap(generateUsernameMd5(keycloakUrl, realm, username), credential);
             CacheFile cacheFile = new CacheFile();
             cacheFile.setCachedData(data);
             mapper.writeValue(file, cacheFile);
-            setOwnerFilePermissions(getCacheFile());
+            setOwnerFilePermissions(file.toPath());
```
dwalluck, line 54: I also changed this to take a `Path` so as to not call `getCacheFile()` a second time.

PR #523, opened 2020-10-19
[NCL-6026] inform user list operations may take a while while running…

jomrazek:
… no filters

### Checklist:


common/src/main/java/org/jboss/pnc/bacon/common/cli/AbstractListCommand.java
```diff
@@ -44,8 +44,13 @@
     @Option(names = "--query", description = "Query parameter (using RSQL)")
     private String query;
 
+    protected boolean print = true;
+
     @Override
     public Integer call() {
+        if (query == null && sort == null && print == true) {
+            System.err.println("Listing entities without filters may take some time, please be patient.");
```
rnc, line 52: Why not use logger at warning level?

jomrazek: sure, that might be better option

PR #524, opened 2020-10-19
Adding post build analysis for artifacts

harshmadhani:
### Checklist:


pig/src/test/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusPostBuildAnalyzerTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void analyzePostBuildDepsFailTest() throws IOException {
+        QuarkusPostBuildAnalyzer quarkusPostBuildAnalyzer = new QuarkusPostBuildAnalyzer(
+                preConfig(),
+                null,
+                "/test-release-path",
+                "/test-extras-path");
+        quarkusPostBuildAnalyzer.trigger();
```
michalszynkiewicz, line 55: it would be good to check the results of the check too, not only that it's doesn't throw an exception

harshmadhani: Added verifyResults to check if the results are valid

PR #526, opened 2020-10-20
NCL-5975 Add gh-pages with documentation

rnc:
### Checklist:


rnc:
Aside : I think the jenkins job should be configured to ignore gh-pages branches.

thescouser89:
Thank you for the PR!!! I unfortunately won't be able to review it soon but I'll try my best to do it by Thursday / Friday. Apologies if this causes any inconvenience, but if it's a blocker, let me know also!

rnc:
> Thank you for the PR!!! I unfortunately won't be able to review it soon but I'll try my best to do it by Thursday / Friday. Apologies if this causes any inconvenience, but if it's a blocker, let me know also!

No problem! :-)

thescouser89:
retest this please

thescouser89:
Thank you for this PR! I'm so torn by this PR :(

On one hand, it looks so nice, is easily accessible and doesn't really depend on any Github technologies like Github wiki. If we'd have to move to a different Git server, we could easily do that.

The reason I'm torn is that Github Wiki has a lower barrier of entry. It's just there on the Wiki tab. I don't know :(

The only request I'd make, if this gets merged, is that we remove the release notes from the Github pages. In all of PNC we consistently have a Changelog wiki and I'd like it to stay that way.

I'll add this as a topic to discuss for the next PiG+PNC CLI meeting.

thescouser89:
What would be your opinion if we deployed the github pages from the master branch? Then the page will be in sync with any PR that has some features.

https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

rnc:
> Thank you for this PR! I'm so torn by this PR :(
> 
> On one hand, it looks so nice, is easily accessible and doesn't really depend on any Github technologies like Github wiki. If we'd have to move to a different Git server, we could easily do that.

Thanks.
 
> The reason I'm torn is that Github Wiki has a lower barrier of entry. It's just there on the Wiki tab. I don't know :(

I disagree slightly :-) I actually think this provides better tracking - we can tie code changes together with a matching PR for documentation. This means we can _review_ documentation changes which isn't possible via the wiki. If people can checkout the repository I don't see the 'barrier' with changing branch (or even having multiple clones) and editing some markdown files - plus we have provided documentation for previewing locally as well. I would also argue that the method to check out the wiki is not that obvious either (though yes you can edit on the gui) 
 
> The only request I'd make, if this gets merged, is that we remove the release notes from the Github pages. In all of PNC we consistently have a Changelog wiki and I'd like it to stay that way.

I wasn't sure on what to do with the changelog so am happy to remove that portion from the gh-pages

As a couple of related points to that:

- I think currently there is duplicate information the the github releases page versus the changelog page - I think it would be better if the former pointed to an appropriate anchor point on the latter? 
- Was there a reason the changelog was chosen to be in a _separate_ git repository rather than a `CHANGELOG.md` sitting alongside the top level `README.md` - this means multiple repository edits rather than having the changelog modification in the _same_ PR ?  

> I'll add this as a topic to discuss for the next PiG+PNC CLI meeting.

Thanks! Is that for 27th October?


> What's your opinion if we deploy the github pages from the master branch? Then the page will be in sync with any PR that has some features, and we also add some documentation to it.
> 
> [docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)

I have never tried that - I'm not convinced that mixing up the jekyll page structure and the code is ideal but I suppose there could be a subfolder. I don't think it makes all the much difference.... dunno ... 


> retest this please

Finally, I think from memory, this won't pass because there isn't a pom in the directory and the CI has not been configured to ignore gh-pages branches.

thescouser89:
We were doing the CHANGELOG.md commited to our repo before. That didn't work because of the constant maintenance of 2 activate branches: master + maintenance one. We switched to the wiki for a single source of truth as a result.

As for the code + documentation together, yup! The documentation / website will be in a subfolder. The advantage I think is that in the same PR you can commit both code + documentation change.

dwalluck:
> As for the code + documentation together, yup! The documentation / website will be in a subfolder. The advantage I think is that in the same PR you can commit both code + documentation change.

I never liked how the gh-pages branch has nothing to do with the code (e.g., comparing it to master would make no sense). Also, it creates a weird situation where you need PRs against both branches simultaneously.

rnc:
> We were doing the CHANGELOG.md commited to our repo before. That didn't work because of the constant maintenance of 2 activate branches: master + maintenance one. We switched to the wiki for a single source of truth as a result.
> 

That makes sense ; its just a shame that github shan't support pr for cloned wiki

> As for the code + documentation together, yup! The documentation / website will be in a subfolder. The advantage I think is that in the same PR you can commit both code + documentation change.

I might need more explanation as I'm not completely convinced - by default i think ( i may be wrong here) github doesn't allow you to publish from multiple branches.  So having docs in multiple branches but only one (the master) is active seems confusing. I think a specific docs branch is less confusing

michalovjan (approved):

PR #538, opened 2020-10-27
Add support for quarkus-bom.json devtools for parent build only in Quarkus Community dep analyser

RotBolt:
cc @michalszynkiewicz 
this much change only should also work for `quarkus-bom`

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
@@ -261,7 +261,8 @@ public class QuarkusCommunityDepAnalyzer extends AddOn {
     }
 
     private String devtoolsJarName() {
-        return "quarkus-product-bom-" + quarkusVersion + ".json";
+        return PigContext.get().getPigConfiguration().getFlow().getRepositoryGeneration().getBomArtifactId() + "-"
+                + quarkusVersion + ".json";
     }
 
     private Set<String> unpackArtifactIdsFrom(Path extensionsPath) {


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -261,7 +261,11 @@ private Path generateQuarkusProject(Predicate<String> artifactIdSelector, String
     }
 
     private String devtoolsJarName() {
-        return "quarkus-product-bom-" + quarkusVersion + ".json";
+        if ((Boolean) getAddOnConfiguration().get("parentOnly")) {
```
michalszynkiewicz, line 264: Let's rename it to "quarkusCoreOnly" or sth like that.

Alternatively, instead of adding a config property, we could check if the `bomArtifactId` for repository generation is `quarkus-bom`

RotBolt: > Alternatively, instead of adding a config property, we could check if the bomArtifactId for repository generation is quarkus-bom

`RepositoryGeneration class` which contains `bomArtifactId`  is not accessible to addons
To make it accessible
- have to make change in `RepoManager`
- have to make change in initialization of addons because no addons take `repositoryGenerationData`

RotBolt: if `RepositoryGenerationsData` is often needed in `addons` then it  would be good idea to make it proper interface to make `RepositoryGenerationData` class accessible

michalszynkiewicz: It should be available via `PigContext.get()`

RotBolt: Nope. 
https://github.com/project-ncl/bacon/blob/37979855adc4436ce89821024f7875516ab6306c/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/PigContext.java#L73

Its `RepositoryData` not `RepositoryGenerationData`. 
repositorydata  contains `Collection<GAV>`, `Collection<File>` and `Path repoPath` only

michalszynkiewicz: `PigContext.pigConfiguration.flow.repositoryConfiguration` sth like that

RotBolt: Have Used bomartifacid now. Thanks for the lead

PR #548, opened 2020-10-30
[NCLSUP-174] Ignore pig-context and target folder

thescouser89:
When generating the hash of the directory config to determine whether
the pig-context.json should be invalidated or not, the hash should
ignore the files 'bacon/pig-context.json' and the contents inside the
'target' folder.

This is to handle a special case where the user may run the `bacon pig`
command using path '.', in which case the 2 files mentioned above will
be created inside the directory config.

Since those 2 files change content all the time, it causes the hash of
the directory to be always different between `bacon pig` runs and
wrongly invalidates the existing pig-context.json because it wrongly
assumes that the *real* content of the directory config has changed.

### Checklist:

### DIFF
--- a/pig/src/test/java/org/jboss/pnc/bacon/pig/impl/utils/HashUtilsTest.java
+++ b/pig/src/test/java/org/jboss/pnc/bacon/pig/impl/utils/HashUtilsTest.java
@@ -59,4 +59,13 @@ class HashUtilsTest {
         assertThat(originalHash).isNotEqualTo(exactCopyHash);
     }
 
+    @Test
+    void shouldIgnoreSomePaths() {
+
+        String originalHash = HashUtils.hashDirectory(testDirs.resolve("original"));
+        String copyIgnoreFileHash = HashUtils.hashDirectory(testDirs.resolve("copy-with-file-to-be-ignored"), path -> {
+            return path.normalize().endsWith("ignore-me");
+        });
+    }
+
 }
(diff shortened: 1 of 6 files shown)


pig/src/test/java/org/jboss/pnc/bacon/pig/impl/utils/HashUtilsTest.java
```diff
@@ -59,4 +59,13 @@ void shouldGiveDifferentHashForAddedFile() {
         assertThat(originalHash).isNotEqualTo(exactCopyHash);
     }
 
+    @Test
+    void shouldIgnoreSomePaths() {
+
+        String originalHash = HashUtils.hashDirectory(testDirs.resolve("original"));
+        String copyIgnoreFileHash = HashUtils.hashDirectory(testDirs.resolve("copy-with-file-to-be-ignored"), path -> {
```
michalszynkiewicz, line 66: how does this test check if the file was ignored?

thescouser89: I'll double check. Maybe I made a booboo again :(

thescouser89: Oh the sha should be different if both were different. Of course I forgot to add the assert line 🤦‍♂️

thescouser89: https://github.com/project-ncl/bacon/pull/558 to fix this. Thanks for pointing it out! :)

PR #552, opened 2020-11-03
[NCL-6182] add a environmentName parameter to build-config.yaml

michalovjan:
### Checklist:

* [ ] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) for your change if user-facing?
* [ ] Have you added unit tests for your change?

PR #553, opened 2020-11-03
[NCL-6182] update gh-pages documentation to include environmentName

michalovjan:
Should be merged after https://github.com/project-ncl/bacon/pull/552

### DIFF
--- a/guide/build-config.md
+++ b/guide/build-config.md
@@ -92,8 +92,9 @@ builds:
 
   scmRevision: <scm revision>
 
-  environmentId: <id> # Either specify environmentId or systemImageId
-  systemImageId: <system image id> # Either specify environmentId or systemImageId
+  environmentId: <id> # Either specify environmentId, environmentName or systemImageId
+  environmentName: <env-name> # Either specify environmentId, environmentName or systemImageId
+  systemImageId: <system image id> # Either specify environmentId, environmentName or systemImageId
 
   dependencies: # Optional: specify which builds need to be done before this one. The build must also be part of the 'builds' list in the YAML
   - <dependency 1 by name>
@@ -138,13 +139,17 @@ builds:
   scmUrl: https://<external git server>/<repo>.git
 ```
 
-#### environmentId v/s systemImageId
+#### environmentId v/s environmentName v/s systemImageId
 
-The build environment used for builds has a unique id and system image id.
+The build environment used for builds has a unique id, name and system image id.
 
-We allow the user to specify the build environment using either `environmentId` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
+We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
 
-The `systemImageId` in contrast, is the same between different PNC servers and should be used if portability is a concern. To get the `environmentId` or the `systemImageId` of the build image, simply run:
+The `environmentName` in contrast, is the same between different PNC servers and should be used if portability is a concern. Environments specified by `environmentName` are always using **the latest updates**, therefore, this is the **preferred** way to specify environments. 
+
+The `systemImageId` is mostly the same between different PNC servers, the difference to `environmentName` is that the `systemImageId` **preserves** the docker-image version of the environment, which means that after some time, the version may become deprecated. (**NOTE**: `systemImageIds` with `:latest` suffix will **no longer be available** in PNC)  
+
+To get the `environmentId`, `environmentName` or the `systemImageId` of the build image, simply run:
 
 ```
 bacon pnc environment list


guide/build-config.md
```diff
@@ -138,13 +139,17 @@ builds:
...
+We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
 
-The `systemImageId` in contrast, is the same between different PNC servers and should be used if portability is a concern. To get the `environmentId` or the `systemImageId` of the build image, simply run:
+The `environmentName` in contrast, is the same between different PNC servers and should be used if portability is a concern. 
+
+The `systemImageId` is mostly the same between different PNC servers, the difference to `environmentName` is that the `systemImageId` **preserves** the docker-image version of the environment whereas environments specified by `environmentName` always use the **latest** updated version (**NOTE**: `systemImageId` with `:latest` suffix behaves the same way as `environmentName`).
```
thescouser89, line 150: `systemImageId` is also now deprecated right @vibe13 ?

michalovjan: Huh. Haven't seen that mentioned anywhere.

thescouser89: Yeah you're right about it not being mentioned anywhere. IIRC from a convo with Andrea, only Marek was using it for CPAAS. If no other users are using it, I think Andrea and I agreed to perhaps deprecate it. I'll wait for his reply on this subject.

michalovjan: After a discussion with Andrea: 
- the fields environmentId and systemImageId won't be deprecated
- environmentName will be highly prefered way for users to specify environments

guide/build-config.md
```diff
@@ -138,13 +139,17 @@ builds:
...
 ```
 
-#### environmentId v/s systemImageId
+#### environmentId v/s environmentName v/s systemImageId
 
-The build environment used for builds has a unique id and system image id.
+The build environment used for builds has a unique id, name and system image id.
 
-We allow the user to specify the build environment using either `environmentId` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
+We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
```
vibe13, line 146: I would add: "Moreover, an environment may become deprecated over time, thus the environmentId should be kept up-to-date, which is not optimal".

guide/build-config.md
```diff
@@ -92,8 +92,9 @@ builds:
 
   scmRevision: <scm revision>
 
-  environmentId: <id> # Either specify environmentId or systemImageId
-  systemImageId: <system image id> # Either specify environmentId or systemImageId
+  environmentId: <id> # Either specify environmentId, environmentName or systemImageId
```
vibe13, line 95: I would modify this section:
# Either specify environmentId, environmentName or systemImageId. If more than one is specified, the priority will be: environmentId, environmentName, systemImageId
environmentId: <id> # The specific id of a build environment, e.g. `20`
systemImageId: <system image id> # The specific builder image name associated with a build environment, e.g. `builder-rhel-7-j8-mvn3.6.0:1.0.0`
environmentName: <env-name> # The name of the builder image, e.g. `OpenJDK 1.8; Mvn 3.6.0`. As this will pick the latest up-to-date version of the image, it's the *recommended option*
  systemImageId: <system image id> # Either specify environmentId, environmentName or systemImageId

guide/build-config.md
```diff
@@ -138,13 +139,17 @@ builds:
...
+We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what we traditionally use, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers.
 
-The `systemImageId` in contrast, is the same between different PNC servers and should be used if portability is a concern. To get the `environmentId` or the `systemImageId` of the build image, simply run:
+The `environmentName` in contrast, is the same between different PNC servers and should be used if portability is a concern. Environments specified by `environmentName` are always using **the latest updates**, therefore, this is the **preferred** way to specify environments.
```
vibe13, line 148: "preferred and recommended" ?

guide/build-config.md
michalovjan: Ack!

guide/build-config.md
michalovjan: I'd still go with preferred as it is more impactful.

guide/build-config.md
michalovjan: Good suggestion. Is this format ok?

vibe13: Argh, not sure how to format it as a yaml comment without being picked from Markdown as a heading. maybe we can move this under the "#### environmentId v/s environmentName v/s systemImageId" heading, in bold?

guide/build-config.md
vibe13: ok makes sense, closing

guide/build-config.md
thescouser89: Gotcha! :)

PR #554, opened 2020-11-05
Add coreOnly config

RotBolt:
cc @michalszynkiewicz

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
@@ -85,6 +85,14 @@ public class QuarkusCommunityDepAnalyzer extends AddOn {
         return NAME;
     }
 
+    private String getBomArtifactId() {
+        return PigContext.get().getPigConfiguration().getFlow().getRepositoryGeneration().getBomArtifactId();
+    }
+
+    private boolean isProductBom(String bomArtifactId) {
+        return bomArtifactId.equals("quarkus-product-bom");
+    }
+
     @Override
     public void trigger() {
         log.info("releasePath: {}, extrasPath: {}, deliverables: {}", releasePath, extrasPath, deliverables);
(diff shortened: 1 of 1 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -314,11 +314,16 @@ private boolean hasQuarkusExtensionMetadata(Path path) {
     private Set<String> gatherProblematicDeps() {
         Set<String> problemmaticDeps = new TreeSet<>();
         problemmaticDeps.addAll(checkBomContents(".*/io/quarkus/quarkus-bom/.*\\.pom"));
-        problemmaticDeps.addAll(checkBomContents(".*/com/redhat/quarkus/quarkus-product-bom/.*\\.pom"));
+        if (Boolean.TRUE.equals(getAddOnConfiguration().get("coreOnly"))) {
```
michalszynkiewicz, line 317: I think that we should reuse the check from https://github.com/project-ncl/bacon/pull/554/files#diff-9b88c7d8548b03f12f77aae69dd7ec253e5b47dc2b53cd73f1bda38d1259765bL269

WDYT @RotBolt ?

RotBolt: I have changed it. Please see it.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -313,12 +317,18 @@ private boolean hasQuarkusExtensionMetadata(Path path) {
 
     private Set<String> gatherProblematicDeps() {
         Set<String> problemmaticDeps = new TreeSet<>();
+        String bomArtifactId = getBomArtifactId();
         problemmaticDeps.addAll(checkBomContents(".*/io/quarkus/quarkus-bom/.*\\.pom"));
-        problemmaticDeps.addAll(checkBomContents(".*/com/redhat/quarkus/quarkus-product-bom/.*\\.pom"));
+        if (bomArtifactId.equals("quarkus-product-bom")) {
```
michalszynkiewicz, line 322: I think it would be a bit better to have the whole check in a separate method too, in case it gets changed. We're doing it in 3 places already now, it would be easy to miss one.

PR #560, opened 2020-11-11
[NCL-6235] update documentation to reflect recent changes to environment handling

michalovjan:
### Checklist:

### DIFF
--- a/guide/build-config.md
+++ b/guide/build-config.md
@@ -93,10 +93,10 @@ builds:
   scmRevision: <scm revision>
   
   # Either specify environmentId, environmentName or systemImageId. 
-  # If more than one is specified, the priority will be: environmentId, environmentName, systemImageId
+  # If more than one is specified, the priority will be: environmentId, systemImageId, environmentName
   environmentId: <id> # The specific id of a build environment, e.g. 20
   environmentName: <env-name> # The name of the builder image, e.g. OpenJDK 1.8; Mvn 3.6.0. As this will pick the latest up-to-date version of the image, it's the recommended option
-  systemImageId: <system image id> # TThe specific builder image name associated with a build environment, e.g. builder-rhel-7-j8-mvn3.6.0:1.0.0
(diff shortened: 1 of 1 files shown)


guide/build-config.md
```diff
@@ -147,9 +147,9 @@ The build environment used for builds has a unique id, name and system image id.
 
 We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`. `environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what has been traditionally used, but it is problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers. Moreover, an environment may become deprecated over time, thus the environmentId should be kept up-to-date, which is not optimal.
```
vibe13, line 148: What about:

We allow the user to specify the build environment using either `environmentId`, `environmentName` or the `systemImageId`.  
`environmentId` refers to the exact id of the build environment as stored in the PNC database. This is what has been traditionally used, but it can be problematic since the ids are different between PNC servers (Devel, Stage, Prod), which makes the `build-config.yaml` not very portable between PNC servers. To overcome the portability issue, it is possible to specify the `systemImageId` instead, which is the same across different PNC servers. Both `environmentId` and `systemImageId` **preserve** the docker-image version of the environment, which means that after some time the version may become deprecated and users may need to update the value from time to time. (**NOTE**: `systemImageIds` with `:latest` suffix will **no longer be available** in PNC).  
The `environmentName` in contrast, is the same between different PNC servers and thus portability across different PNC servers is not a concern. However the big difference is that environments specified by `environmentName` are always using **the latest released docker image**. This is important because there will be changes from time to time to the build configurations in PNC, which will cause **rebuilds** of the affected components.

michalovjan: I like it :)

thescouser89 (approved):

PR #579, opened 2020-12-07
Reverting jaxb version to 2.2.11 due to failures while running QuarkusCommunityDepAnalyzer add-on

harshmadhani:
### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
 
     <properties>
         <compiler.plugin.version>3.8.1</compiler.plugin.version>
-        <jaxb.version>3.0.0</jaxb.version>
+        <jaxb.version>2.2.11</jaxb.version>
         <junit.version>5.7.0</junit.version>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>


michalszynkiewicz:
@harshmadhani do you know who did the problemmatic update? Maybe it's worth adding t his person as a reviewer, in case there was some specific reason to update.

rnc:
It was that dependabot tool ( https://github.com/project-ncl/bacon/commit/e87ee7eac1c69858393e1cd1836578a7a944941b ) - I therefore think this change *must* have a test so any future change of this version is caught via a test failure.

rnc:
@harshmadhani @michalovjan @thescouser89 I am also seeing 
```
Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/xml/bind/v2/model/annotation/AnnotationReader
```
It seems that same PR is the cause. Is there a reason we need this dependabot running?

michalszynkiewicz:
Let's merge it, @harshmadhani @RotBolt could you guys create a test for this add-on (if we don't have it already) in a separate PR?

RotBolt:
Sure, we will create a test in next PR for this issue

PR #602, opened 2021-01-15
Stop source artifact downloads from retrying

AntStephenson:
Only try once to download source artifacts as they can be missing (and from our repo there are a lot of missing source artifacts) which can cause the repo build time to be considerably larger. 

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
@@ -50,14 +50,23 @@ public class ExternalArtifactDownloader {
         String indyUrl = gav.isTemporary() ? Indy.getIndyTempUrl() : Indy.getIndyUrl();
 
         URI downloadUrl = URI.create(String.format("%s/%s", indyUrl, gav.toUri()));
-        try {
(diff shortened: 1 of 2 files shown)


rnc:
Shouldn't source artifacts always exist i.e. its rare they won't exist as the Gates mandate they exist? So shouldn't the repository be fixed instead?

AntStephenson:
@rnc This is not the reality we see unfortunately - especially made worse when producing a BUILD_GROUP repository with all build deps too, there are quite a lot of third parties that do not have source jars available.

rnc:
> @rnc This is not the reality we see unfortunately - especially made worse when producing a BUILD_GROUP repository with all build deps too, there are quite a lot of third parties that do not have source jars available.
> 
> Edit: Just to also point out by default source jars are optional and do not fail a Bacon Maven repository build probably for this reason exactly.

We should probably take this offline as its separate to this pr (no objections per se) but i thought the gates mandated source jars for every artifact i think?

AntStephenson:
@rnc Happy to take it offline, can talk more about it on Monday.

thescouser89 (approved):
LGTM!

thescouser89:
I'll merge this PR.

Thanks @AntStephenson for blessing us with your first contribution! Much appreciated :)

Please let me know also about the outcome of the discussion and we'll see!

PR #608, opened 2021-01-20
[NCLSUP-244] release: only generate script+nvr list if data present

thescouser89:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
@@ -165,7 +165,7 @@ public class Pig {
     }
 
     /* System.setProperty("java.util.concurrent.ForkJoinPool.common.parallelism", "10"); */
-    @Command(name = "run", description = "Run all the steps")
+    @Command(name = "run", description = "Run all the steps, except the release & pre-process-yaml phases")
     public static class Run extends PigCommand<PigRunOutput> {
 
         // TODO: it is doable to do this step with build group id only, add this functionality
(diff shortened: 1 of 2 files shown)


rnc:
Sorry to be picky but when the CLI has: 
```
Commands:
  configure         Configure PNC entities
  build             Build
  run               Run all the steps
  repo              GenerateRepository
  licenses          GenerateLicenses
  javadocs          GenerateJavadocs
  sources           GenerateSources
  shared-content    GenerateSharedContentAnalysis
  docs              GenerateDocuments
  release           Push builds to brew, generate the NVR list, close the PNC milestone, generate the upload to candidates script
  pre-process-yaml  Show the final YAML content with variables injected
  addons            Addons
```
does 'run all the phases' include configure / preprocess / addons ? I am _guessing_ yes to the latter but no to the former two but I'm not sure :-)

thescouser89:
That is a good question! Not pre-process-yaml, since I guess it's not really a step (but yeah how many users will know that!), but yes for configure and addons.

rnc:
> That is a good question! Not pre-process-yaml, since I guess it's not really a step (but yeah how many users will know that!), but yes for configure and addons.

Ah so I was wrong :-) So maybe it needs to be "run all the steps i.e. all other commands listed except release & pre-process-yaml" ?

thescouser89:
Yeah makes sense to me. I'll edit my pr to include pre-process-yaml also in the description. Thanks Nick for reviewing it and your suggestion! :)

rnc (approved):
Cool , I think that makes more sense. 
If we were being _really_ picky I think it should be `release & pre-process` to make the grammar flow better :-)

thescouser89:
sure!

PR #609, opened 2021-01-20
Update docs with some repo generation bits & offliner manifest generation

AntStephenson:
Added some docs about generating a manifest for the offliner tool & docs for including or excluding artifacts from the repository generation phase. I see there are a bunch of different strategies in PiG that I don't use so I've just added a foundation for them to be explained by someone else.

### DIFF
--- a/guide/pig.md
+++ b/guide/pig.md
@@ -86,12 +86,73 @@ This section covers the required deliverables for a product release:
 </tr>
 </table>
 
-## Add ons
+### Maven Repository
+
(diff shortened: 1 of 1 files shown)


rnc:
Thanks for the contribution! I'll let @thescouser89  review but I do see that Dustin added some things about repos in https://project-ncl.github.io/bacon/guide/build-config.html#maven-repository-generation ... not sure of the best structure to be honest.

thescouser89:
Thanks for the PR @AntStephenson ! LGTM!

Yeah we'll have to unify those 2 docs Nick. I'll check and maybe create another PR to unify it. Thanks also for your review! :)

PR #612, opened 2021-01-26
[NCL-6328] Review regexps in rh-license-exceptions.json

thescouser89:
### Checklist:

### DIFF
--- a/pig/src/main/resources/rh-license-exceptions.json
+++ b/pig/src/main/resources/rh-license-exceptions.json
@@ -43,7 +43,7 @@
   {
     "groupId": "antlr",
     "artifactId": "antlr",
-    "version-regexp": "2.*",
+    "version-regexp": "2\\..+",
     "licenses": [
       {
         "name": "The Antlr 2.7.7 License",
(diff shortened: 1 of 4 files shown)


pig/src/main/resources/rh-license-exceptions.json
```diff
@@ -794,7 +794,7 @@
   {
     "groupId": "io.agroal",
     "artifactId": "agroal-api",
-    "version-regexp": "1.8*",
+    "version-regexp": "1\\.8*",
```
michalszynkiewicz, line 797: this looks a bit suspicious. Shouldn't it be sth like `1\\.8.*` ? (just a note, this change doesn't actually change the suspicious part)

michalszynkiewicz (approved):

pig/src/main/resources/rh-license-exceptions.json
thescouser89: I'll have a look. Thanks for pointing this out!

thescouser89: agroal actually has version `1.8`, `1.9` etc.

`1\\.8.*` makes sense in that optic. Will do the change!

pkocandr: I would probably go for `1\\.8(\\..*)?` to not include e.g. 1.80 in a few years. WDYT?

PR #613, opened 2021-01-26
[MMENG-1127] Trust a 404 and don't retry

thescouser89:
When we get HTTP response 404 from our HTTP client, don't do further
retries in the hopes that the 404 changes to a 200 status.

CC: @AntStephenson 

### Checklist:


AntStephenson (commented):
Looks better!

PR #617, opened 2021-01-31
Skipping tests while building temporary project

harshmadhani:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
@@ -148,13 +148,20 @@ public class QuarkusCommunityDepAnalyzer extends AddOn {
     }
 
     private Path buildReactiveProject(String settingsSelector) {
-        Path projectPath = generateQuarkusProject(artifactId -> artifactId.contains("reactive"), settingsSelector);
+        Path projectPath = generateQuarkusProject(
+                artifactId -> artifactId.contains("micrometer"),
+                artifactId -> !artifactId.contains("metrics"),
+                settingsSelector);
         buildProject(projectPath, settingsSelector);
         return projectPath;
     }
 
     private Path buildNonReactiveProject(String settingsSelector) {
-        Path projectPath = generateQuarkusProject(artifactId -> !artifactId.contains("reactive"), settingsSelector);
+        Path projectPath = generateQuarkusProject(
+                artifactId -> !artifactId.contains("micrometer"),
+                artifactId -> artifactId.contains("metrics"),
+                settingsSelector);
+        buildProject(projectPath, settingsSelector);
         buildProject(projectPath, settingsSelector);
         return projectPath;
     }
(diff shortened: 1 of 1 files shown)


harshmadhani:
cc @RotBolt @paulgallagher75 @michalszynkiewicz

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -271,7 +273,7 @@ private Path generateQuarkusProject(Predicate<String> artifactIdSelector, String
     private String devtoolsJarName() {
         String bomArtifactId = getBomArtifactId();
         if (!isProductBom(bomArtifactId)) {
-            return "quarkus-bom-descriptor-json-" + quarkusVersion + ".json";
+            return "quarkus-bom-quarkus-platform-descriptor-" + quarkusVersion + "-" + quarkusVersion + ".json";
```
RotBolt, line 276: I think now the name of `bom-desciptor-json` will depend on artifact now 
it will be of form 
<bom-artifact-id>-quarkus-platform-descriptor-<quarkusVersion>-<quarkusVersion>.json
we need to include <bom-artifact-id> now in this

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -231,7 +233,7 @@ private Path generateQuarkusProject(Predicate<String> artifactIdSelector, String
                 .filter(artifactIdSelector)
                 .collect(Collectors.toList());
         String command = String.format(
-                "mvn -X io.quarkus:quarkus-maven-plugin:%s:create -DprojectGroupId=tmp -DprojectArtifactId=tmp "
+                "mvn -X -Dmaven.test.skip=true io.quarkus:quarkus-maven-plugin:%s:create -DprojectGroupId=tmp -DprojectArtifactId=tmp "
```
michalszynkiewicz, line 236: `-Dmaven.test.skip=true` seems unnecessary here

RotBolt: I think it's necessary otherwise project creation fails. 
`quarkus-kafka-streams-extension` require specific config in application.properties file which is not there at the time of project creation. 

And it was actually test which was failing. So skipping tests actually solves the issue for us.

michalszynkiewicz: are you sure it's `:create` goal that fails? That would mean the goal is unusable

RotBolt: yes thats the goal that fails

harshmadhani: The create goal works without skipping tests. We required to skip tests only building the project. I tested it locally and it works

RotBolt: wierd it failed on my local :\

harshmadhani: Done !

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
```diff
@@ -148,13 +148,20 @@ public void trigger() {
     }
 
     private Path buildReactiveProject(String settingsSelector) {
-        Path projectPath = generateQuarkusProject(artifactId -> artifactId.contains("reactive"), settingsSelector);
+        Path projectPath = generateQuarkusProject(
+                artifactId -> artifactId.contains("micrometer"),
+                artifactId -> !artifactId.contains("reactive,metrics"),
```
michalszynkiewicz, line 153: should it be a check that  artifact id contains either `reactive` or `metrics`, not `reactive,metrics`?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
harshmadhani: But this is just to find the json file and not else. Would this still affect us?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/QuarkusCommunityDepAnalyzer.java
harshmadhani: Yes, probably we can remove reactive and keep only micrometer in one condition and metrics in another, wdyt?

PR #630, opened 2021-02-19
[NCLSUP-264] Reduce bacon logging

thescouser89:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
@@ -54,7 +54,7 @@ public class ExternalArtifactDownloader {
             FileDownloadUtils.downloadTo(downloadUrl, targetPath);
         } catch (RuntimeException any) {
             if (sourcesOptional && "sources".equals(gav.getClassifier()) || "javadoc".equals(gav.getClassifier())) {
-                log.warn("Unable to download sources for {}", gav, any);
+                log.warn("Unable to download sources for {}: {}", gav, any.getMessage());
             } else {
                 throw any;
             }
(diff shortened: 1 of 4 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileDownloadUtils.java
```diff
@@ -61,7 +61,6 @@
     public static void downloadTo(URI downloadUrl, File targetPath) {
         log.info("Downloading {} to {}", downloadUrl, targetPath);
```
jomrazek, line 62: maybe put this as debug? user is still able to debug with --verbose

thescouser89: true true. will change it to debug

thescouser89:
@codecov how was your weekend?

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
```diff
@@ -54,7 +54,7 @@ public static File downloadExternalArtifact(GAV gav, File targetPath, boolean so
             FileDownloadUtils.downloadTo(downloadUrl, targetPath);
         } catch (RuntimeException any) {
             if (sourcesOptional && "sources".equals(gav.getClassifier()) || "javadoc".equals(gav.getClassifier())) {
-                log.warn("Unable to download sources for {}", gav, any);
+                log.warn("Unable to download sources for {}: {}", gav, any.getMessage());
```
jomrazek, line 57: just spoke about this with Nick, with your change output would be smaller so I would leave it for now, but I don't know how many times it could be printed to user and if it's that important so :see_no_evil: . At the end I would see bacon informing users just about major steps in process and errors. If users want detailed logs they could switch to verbose output

jomrazek:
LGTM :)

jomrazek (approved):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/ExternalArtifactDownloader.java
thescouser89: Yeah we can adjust later in another commit. We'll learn as we go! :)

PR #632, opened 2021-02-22
Removing artifacts which contained invalid license name

harshmadhani:
Removing licenses which contained invalid name "Lesser General Public License v3.0 or later". As the same artifacts are present under correct license names in the json file, I removed the license instead of just updating the license name 

### Checklist:


RotBolt:
looks fine to me

PR #643, opened 2021-03-16
[NCLSUP-298] use temporary hosted repo instead of a group

michalovjan:
The temporary builds group consists of only the hosted repo so it doesn't make sense not to use the hosted repo directly.

### Checklist:


jomrazek (approved):
lgtm

PR #647, opened 2021-03-24
[NCLSUP-305] Document releaseStorageUrl usage in bacon/pig documentation

jomrazek:
### Checklist:

### DIFF
--- a/guide/build-config.md
+++ b/guide/build-config.md
@@ -66,7 +66,7 @@ product:
 ```
 The `version` specifies the product version we are building for, and the `milestone` represents the 'step' at which we are in the version. A typical milestone value is DR1, DR2, etc or ER1, ER2 etc. The combination of the version and milestone forms the full release name of the product: e.g 1.0.0.CR1.
 
-**TODO**: talk about automatic generation of milestone versions with `--releaseStorageUrl`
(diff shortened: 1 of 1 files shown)


guide/build-config.md
```diff
@@ -66,7 +66,7 @@ product:
 ```
 The `version` specifies the product version we are building for, and the `milestone` represents the 'step' at which we are in the version. A typical milestone value is DR1, DR2, etc or ER1, ER2 etc. The combination of the version and milestone forms the full release name of the product: e.g 1.0.0.CR1.
 
-**TODO**: talk about automatic generation of milestone versions with `--releaseStorageUrl`
+The `--releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. Using '*' in your milestone you could also use it for automatic incrementation of milestone value. For example using 'ER*' will check release storage and increment version to next available milestone version.
```
rnc, line 70: I think we should add when it is used (i.e. phase). If '*' isn't used in the milestone, what else is it used for ? (as you said `also` )

rnc: Sorry , but I still don't find that clear:
1. In `release` I didn't think it does any actual uploading to rcm-guest? i.e. therefore _how_ is it used in `release` stage? - does it generate and alter scripts ?
2. What phase is milestone increment done in? I presume `run`, but its not explicitly said.

In other words, I think its needs to be made explicit how its used in the different phases please.

rnc (changes requested):
See inline comments.

guide/build-config.md
jomrazek: "upload to candidates script" is term from pig release documentation so I used that, I can put it in quotes or use "the"

rnc (approved):

thescouser89 (approved):
LGTM

PR #649, opened 2021-03-30
[NCL-6473] Bacon should improving reports for build groups

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
@@ -17,6 +17,8 @@
  */
 package org.jboss.pnc.bacon.pig.impl.pnc;
 
+import org.jboss.pnc.bacon.common.exception.FatalException;
+import org.jboss.pnc.bacon.config.Config;
 import org.jboss.pnc.bacon.pig.impl.utils.SleepUtils;
 import org.jboss.pnc.client.ClientException;
 import org.jboss.pnc.client.GroupBuildClient;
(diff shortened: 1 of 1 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
```diff
@@ -112,9 +115,12 @@ boolean isSuccessfullyFinished(String groupBuildId) {
                 case SUCCESS:
                     return verifyAllBuildsInGroupBuildInFinalStateWithProperCount(groupBuildId);
                 default:
-                    throw new RuntimeException("Build group failed " + groupBuild);
+                    throw new RuntimeException(
+                            "Build group failed " + new URL(
```
rnc, line 119: Why use `new URL`? Also isn't it `group-builds` not `group-build`? 
Why not make it 
```
throw new FatalException("Build group failed http://{}/pnc-web/#/group-builds/{}",
		 Config.instance().getActiveProfile().getPnc().getUrl(), groupBuildId)
```

rnc (changes requested):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
jomrazek: Made it url so it's clickable but it's working without it too so you're right. For formatting I think it's more readable my way but I don't mind changing in, will send fix.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
```diff
@@ -112,7 +114,10 @@ boolean isSuccessfullyFinished(String groupBuildId) {
                 case SUCCESS:
                     return verifyAllBuildsInGroupBuildInFinalStateWithProperCount(groupBuildId);
                 default:
-                    throw new RuntimeException("Build group failed " + groupBuild);
+                    throw new FatalException(
+                            "Build group failed {}/pnc-web/#/group-builds/{}",
+                            Config.instance().getActiveProfile().getPnc().getUrl(),
```
rnc, line 119: Do you need `http://` on it?
Should this have a test?

jomrazek: http:// is part of the pnc url so not needed here, as for test I don't think so, if url isn't correct it fails earlier in gathering of configuration part and this is just change of the printed info.
Screenshot from 2021-03-30 14-04-46

rnc: Ok cool

rnc (approved):

PR #651, opened 2021-03-31
[NCL-6464] Deprecate Timestamped teporary builds option in Bacon

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
@@ -82,10 +82,12 @@ public class PncBuilder implements Closeable {
             boolean tempBuildTS,
             RebuildMode rebuildMode) {
         log.info("Performing builds of build group {} in PNC", group.getId());
+        if (tempBuildTS)
+            log.warn(
+                    "Temporary builds with timestamp alignment are not supported, running temporary builds instead...");
         GroupBuildParameters buildParams = new GroupBuildParameters();
         buildParams.setRebuildMode(rebuildMode);
         buildParams.setTemporaryBuild(tempBuild);
-        buildParams.setTimestampAlignment(tempBuildTS);
         GroupBuildRequest request = GroupBuildRequest.builder().build();
 
         try {
--- a/pnc/src/main/java/org/jboss/pnc/bacon/pnc/BuildCli.java
+++ b/pnc/src/main/java/org/jboss/pnc/bacon/pnc/BuildCli.java
@@ -93,7 +93,9 @@ public class BuildCli {
         private String rebuildMode;
         @Option(names = "--keep-pod-on-failure", description = "Keep the builder pod online after the build fails.")
         private boolean keepPodOnFailure = false;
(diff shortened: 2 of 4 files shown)


pnc/src/main/java/org/jboss/pnc/bacon/pnc/BuildCli.java
```diff
@@ -93,7 +93,7 @@
         private String rebuildMode;
         @Option(names = "--keep-pod-on-failure", description = "Keep the builder pod online after the build fails.")
         private boolean keepPodOnFailure = false;
-        @Option(names = "--timestamp-alignment", description = "Do timestamp alignment with temprary build.")
+        @Option(names = "--timestamp-alignment", description = "DEPRECATED!!! - Do timestamp alignment with temprary build.")
```
vibe13, line 96: "temporary" :-)

jomrazek: fixed!

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
```diff
@@ -82,10 +82,11 @@ private GroupBuild run(
             boolean tempBuildTS,
             RebuildMode rebuildMode) {
         log.info("Performing builds of build group {} in PNC", group.getId());
+        if (tempBuildTS)
+            log.info("Temporary builds with timestamp alignment are deprecated, running temporary build...");
```
rnc, line 86: Shouldn't this be a warning ? :-) Does this mean its running without timestamp alignment and just running a temporary build?

jomrazek: Yes it means exactly that, I guess it might be also warning but I believe we didn't receive feedback somebody is actually using it so I went with info. For deprecated, is no longer supported better?

rnc: My understanding of deprecated is that e.g. "functionality x will be removed in a future version so migrate to the replacement y".
So yes, I agree and think  "Temporary builds with timestamp alignment is not supported ; running temporary build instead..." is  better.

In terms of info/warn ; personally I think if someone picks an option that you are trying to alert them that is no longer functions, warning is better. 

 @vibe13  - thought?

pnc/src/main/java/org/jboss/pnc/bacon/pnc/BuildCli.java
```diff
@@ -93,7 +93,9 @@
         private String rebuildMode;
         @Option(names = "--keep-pod-on-failure", description = "Keep the builder pod online after the build fails.")
         private boolean keepPodOnFailure = false;
-        @Option(names = "--timestamp-alignment", description = "Do timestamp alignment with temprary build.")
+        @Option(
+                names = "--timestamp-alignment",
+                description = "DEPRECATED!!! - Do timestamp alignment with temporary build.")
```
rnc, line 98: If its no longer supported surely this shouldn't say deprecated?

thescouser89 (approved):

PR #648, opened 2021-03-26
Automate creation of vertx artifacts list which are used in quarkus

SaumyaSingh1:
CC @RotBolt @harshmadhani 
This is a draft PR for "Automate the creation of vertx artifacts list which are used in quarkus".

PR #659, opened 2021-04-09
Adding details of vertx addon in the doc

SaumyaSingh1:
Hi @thescouser89 @rnc,

Here is the PR updating the doc with the newly added add-on  `Vertx Artifact Finder` #648  :)

I previewed the changes using ` Jekyll ` and it is looking like this - 

image


thescouser89:
LGTM! Thanks for the changes and taking care of the docs also! :hugs:

PR #660, opened 2021-04-12
WIP : Adding JUnit for Vertx Addon

saumya1singh:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/quarkus/VertxArtifactFinder.java
```diff
@@ -63,7 +63,7 @@ public VertxArtifactFinder(
             String extrasPath,
             Path repoPath) {
         this(pigConfiguration, builds, releasePath, extrasPath);
-        repoPath = this.repoPath;
+        this.repoPath = repoPath;
```
thescouser89, line 66: good catch!

PR #684, opened 2021-04-29
[NCLSUP-306] Disable CLI output with --no-color flag

thescouser89:
When running bacon in specific environment where the logs are piped into
files, the ANSI color commands show up in those logs and make them hard
to read.

This commit adds the --no-color flag to disable color/ANSI output in
picocli output. The user can also disable coloring by setting the
'NO_COLOR' environment variable with any value (https://no-color.org/).

We also remove the '%highlight' in the logback file to disable coloring
of the message level printed in stderr for logging.

ansi
ansi-log

### Checklist:

### DIFF
--- a/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
+++ b/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
@@ -113,7 +113,31 @@ public class App {
         this.configPath = configPath;
     }
 
+    /**
+     * Disable color in Picocli output
+     *
+     * @param noColorPresent if no-color mode is activated
+     */
+    @Option(
+            names = { "--no-color" },
+            description = "Disable color output. Useful when running in a non-ANSI environment",
+            scope = INHERIT)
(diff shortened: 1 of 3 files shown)


dwalluck:
I was fairly certain that picocli already supposed setting this env var, so it really wouldn't require any code changes.

thescouser89:
Oh cool. I'll remove my check for NO_COLOR. I believe however the flag is required if the user wants to explicitly specify "I want no color" without having to set any env vars.

dwalluck:
> Oh cool. I'll remove my check for NO_COLOR. I believe however the flag is required if the user wants to explicitly specify "I want no color" without having to set any env vars.

Yes, I figured maybe you wanted to do what it does with a flag instead of property or env var, but you didn't need additional logic to check for it. Sorry if I wasn't clear.

thescouser89:
No worries! Thanks for pointing out about the `NO_COLOR` one. I didn't read far enough in the docs. Thank you! :)

dwalluck (approved):

PR #686, opened 2021-04-29
[NCL-6058] Add quiet option (-q)

thescouser89:
When the quiet option is enabled, the logging level is set to 'ERROR' (no
logs, unless critical will be printed) and any YAML/JSON printed at the end will also not
be printed

### Checklist:

### DIFF
--- a/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
+++ b/cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
@@ -100,6 +100,17 @@ public class App {
         }
     }
 
+    /**
+     * Make logback quiet if quiet flag is set
+     */
+    @Option(names = { "-q", "--quiet" }, description = "Silent output", scope = INHERIT)
+    public static void setQuietIfPresent(boolean quiet) {
+
+        if (quiet) {
+            ObjectHelper.setRootLoggingLevel(ObjectHelper.LOG_LEVEL_SILENT);
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


thescouser89:
Wait for https://github.com/project-ncl/bacon/pull/684 to be merged. Reason is that tests are picky about the cli output, and with this commit it changes that.

cli/src/main/java/org/jboss/pnc/bacon/cli/App.java
```diff
@@ -100,6 +100,17 @@ public static void setVerbosityIfPresent(boolean verbose) {
...
     }
 
+    /**
+     * Make logback quiet if quiet flag is set
+     */
+    @Option(names = { "-q", "--quiet" }, description = "Silent output", scope = INHERIT)
+    public static void setQuietIfPresent(boolean quiet) {
+
+        if (quiet) {
+            ObjectHelper.setRootLoggingLevel(Level.OFF);
```
rnc, line 110: What about setting it to Level.WARN ? Or even Level.ERROR ? (which matches what Maven does) i.e. disable INFO logging to be 'quiet? I think turning everything off is a bit too extreme?

thescouser89: quiet is quiet :)

I guess Level.ERROR sounds about right.

thescouser89: Thanks Nick for the input! I changed it to ERROR now. Good call! :)

PR #692, opened 2021-05-05
[NCL-6505] Add ability to specify brewPullActive in Bacon

jomrazek:
### Checklist:


jomrazek:
If you have better description in mind I'm looking forward to change this one. I was thinking about adding flag for pig part but I don't consider this as parameter that might be different every build so I didn't add it and could be used just as parameter in build-config.yaml

thescouser89:
Description from our internal docs:

```
Since PNC is the main build system for Middleware product Maven builds, the number of dependencies in Brew used in current PNC builds is pretty low and the feature slows down both alignment and the build itself.

This flag allows the user to specify whether to activate the brew pull for specific build configs.
```

Credit for the doc: @pkocandr

thescouser89 (approved):
LGTM! :eyes:

PR #766, opened 2021-08-17
Add `pnc builg get-align-logs` command

janinko:
### Checklist:


pnc/src/main/java/org/jboss/pnc/bacon/pnc/BuildCli.java
```diff
@@ -303,6 +303,38 @@ public Integer call() throws Exception {
         }
     }
 
+    @Command(name = "get-align-log", description = "Get alignment log.")
```
jomrazek, line 306: you need to add it as subcommand on top of the class too

jomrazek (approved):
lgtm

PR #767, opened 2021-08-19
Update the list of extensions based for Quarkus 2.2

SaumyaSingh1:
Artifacts/Extensions Updated. 
cc @harshmadhani


harshmadhani:
LGTM !

harshmadhani:
Can we merge it @paulgallagher75 @thescouser89 ?

paulgallagher75 (commented):
So this list looks like it has always been a part of the Quarkus addon but wouldn't it be better to break the info out of Bacon and put into a separate file then the update to source don't have to go thru here, I will merge as it needs to be done however for the future I think not having the list hard-coded here is better

harshmadhani:
@paulgallagher75 agree on this. Its better to be a separate file and just to be able to pass the path of that file as a parameter for DepAnalyzer to work

PR #770, opened 2021-08-19
[NCLSUP-425] Add retries for Keycloak client

thescouser89:
The first commit ports from the old Unirest 1.x to Unirest 3.11.10, which comes with package name changes and a different way to configure the object mapper. Overall, Unirest 3.11.10 has a simpler design, which I like.

The second commit attempts to solve NCLSUP-425, which is to add retries to the Keycloak client. We have retries when reaching PNC-Orch, and retries when downloading from Indy, so there should also be a retry for the Keyclaok client! :) The retries however should help us resolve some instability that we've been seeing when reaching the SSO server lately.

### Checklist:

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -31,24 +31,8 @@
 
         <!-- unirest and family -->
         <dependency>
-            <groupId>com.mashape.unirest</groupId>
+            <groupId>com.konghq</groupId>
             <artifactId>unirest-java</artifactId>
-        </dependency>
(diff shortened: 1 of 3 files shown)


rnc:
For reference, PME ported to use Unirest 3.x a while back as well

jomrazek (approved):
Nice! lgtm

thescouser89:
Thanks for the review! :)

PR #777, opened 2021-08-26
[NCL-5983] Deprecate external scm url

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
@@ -74,6 +74,7 @@ public class BuildConfig {
     private @NotBlank String project;
     private @NotBlank String buildScript;
     private String scmUrl;
+    private String externalScmUrl;
     private @NotBlank String scmRevision;
     private String description;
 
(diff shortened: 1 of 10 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
```diff
@@ -235,8 +235,7 @@ public boolean isTheSameAs(BuildConfiguration old, boolean skipBranchCheck, bool
     }
 
     private boolean urlsEqual(SCMRepository repo) {
-        return StringUtils.equals(externalScmUrl, repo.getExternalUrl())
-                || StringUtils.equals(scmUrl, repo.getInternalUrl());
+        return StringUtils.equals(scmUrl, repo.getInternalUrl());
```
michalovjan, line 238: `scmUrl` field can also be external right?

michalovjan (changes requested):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/BuildConfig.java
jomrazek: yup :) you can put external url e.g. github to scmUrl field

michalovjan: sure

michalovjan (approved):

PR #800, opened 2021-09-24
Fix license  : zstd-jni

SaumyaSingh1:


SaumyaSingh1:
@paulgallagher75 please merge :)
Thanks.

paulgallagher75:
I will merge this time however you should use `"version-regexp"` to remove the need to add an entry for every redhat version, the license will not change on the community patch version either so its safe to cover something like `1\\.4\\.9\\.[0-9]+-redhat-*` (this is just a quick guess to get this right test what I just put down first)

PR #804, opened 2021-09-24
[NCL-6720] Created command to list delivered artifacts and bugfix

jakubvanko:
### Checklist:


janinko (changes requested):
Overall looks good.
Maybe since we use 2.2 api now, the bacon version should be changed to 2.2.0-SNAPSHOT? WDYT @jomrazek ?

jomrazek:
> Overall looks good.
> Maybe since we use 2.2 api now, the bacon version should be changed to 2.2.0-SNAPSHOT? WDYT @jomrazek ?

Not sure we're tied to that but next version will be 2.2.x as we're doing bigger changes as adding DA commands

janinko (approved):

PR #904, opened 2022-01-25
[NCL-5981] Add ability to provide a suffix for deliverables

thescouser89:
In NCL-5981, the user would like to provide a suffix for the
deliverables files and folder inside the zips.

For example: <product>-<version>-<suffix>-maven-repository.zip

This commit adds an additional optional field in the build-config.yaml named
'outputSuffix' that when specified, will be used as the suffix.

Example:
```
outputSuffix: picard
```

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/PigContext.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/PigContext.java
@@ -115,7 +115,14 @@ public class PigContext {
         initFullVersion(pigConfiguration, releaseStorageUrl);
 
         if (deliverables == null) {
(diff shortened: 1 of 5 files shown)


thescouser89:
@jomrazek  Let me know what you think about this or if you think we should do some changes about it! :)

jomrazek:
Looks great! Maybe we can add test with stage, version etc. to not compare just "-"?

jomrazek (approved):

thescouser89:
Yup good point! I'll add those

thescouser89:
I'll create a PR for doc to document this.

thescouser89:
Document: #905

PR #909, opened 2022-02-09
Make excludeArtifacts for resolve only resolution

thescouser89:
The `excludeArtifacts` in the maven repository section for PiG works for
most maven repository generation except for RESOLVE_ONLY. This commit
adds the ability to use `excludeArtifacts` with RESOLVE_ONLY

The implementation works by first generating the final maven repository
from the bom specified, then deleting the artifacts in
`excludeArtifacts` from that maven repository.

The artifacts are specified in "Maven Identifier" format (aka
`<group-id>:<artifact-id>:<packaging>:<version>`). We convert that Maven
identifier into a filesystem path inside the maven repository folder and
delete it. Regexes are supported.

The other option was to generate the maven repository with the excluded
artifacts specified as exclusions. But I couldn't find any way to do
that using the Maven Resolver object.

### Checklist:


thescouser89:
CC: @harshmadhani @saumya1singh

jomrazek (approved):
LGTM :+1:

thescouser89:
Thanks for the review @jomrazek ! :)

PR #912, opened 2022-03-02
Add camel addon for build-from-source statistics and info

cunningt:
Add camel addon for build-from-source statistics and info


thescouser89:
Thanks for porting this add-on to PiG! :)

PR #913, opened 2022-03-03
Fix check for correct gav formatting

jomrazek:
https://project-ncl.github.io/bacon/guide/build-config.html#maven-repository-generation G:A:V is required but we're testing if at least 2 things are present, creating out of bounds error when user wrongly formatted file

### Checklist:


thescouser89:
Good catch!

PR #917, opened 2022-03-08
[NCLSUP-609] Improve pig cancel behavior

jomrazek:
Previous message was confusing in terms why is it shown during cancel, also as pnc allow multiple running group builds of one group configuration I think we should just inform user and don't need to throw error.

context from NCLSUP-609:
Before cancel we're checking if more than one instance of GroupBuild with same GroupConfig ID is running, because we're getting GroupBuild by name used in build-config.yaml and not by ID. PNC is allowing multiple GroupBuilds of same GroupConfig running and that could cause this problem where we can't decide which one to cancel just using the name from build-config.yaml although it's not recommended to do this using bacon.


thescouser89 (approved):
LGTM!

PR #921, opened 2022-03-23
Move build to source statistics to their own log, change file name of  dependencies to align to something more generic

cunningt:
Move build to source statistics to a separate logging file, change file name of  dependencies to align to something more generic

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/camel/RuntimeDependenciesToAlignTree.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/camel/RuntimeDependenciesToAlignTree.java
@@ -65,15 +65,21 @@ public class RuntimeDependenciesToAlignTree extends AddOn {
 
     @Override
     public void trigger() {
-        String filename = extrasPath + "CamelRuntimeDependenciesToAlignTree.txt";
-        log.info("Running RuntimeDependenciesToAlignTree - report is {}", filename);
-        try (PrintWriter file = new PrintWriter(filename, StandardCharsets.UTF_8.name())) {
+        String filename = extrasPath + "DependenciesToAlignTree.txt";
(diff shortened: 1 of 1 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/camel/RuntimeDependenciesToAlignTree.java
```diff
@@ -65,15 +65,21 @@ private String parseDependency(String line) {
...
     public void trigger() {
-        String filename = extrasPath + "CamelRuntimeDependenciesToAlignTree.txt";
-        log.info("Running RuntimeDependenciesToAlignTree - report is {}", filename);
-        try (PrintWriter file = new PrintWriter(filename, StandardCharsets.UTF_8.name())) {
+        String filename = extrasPath + "DependenciesToAlignTree.txt";
+        String buildFromSourceStatsFileName = extrasPath + "BuildFromSourceStats.txt";
+        log.info("Running DependenciesToAlignTree - report is {}", filename);
+        PrintWriter file = null;
+        PrintWriter buildFromSourceStatsFile = null;
+        try {
```
thescouser89, line 73: Hi @cunningt ! Could this code be refactored to use multiple items in the [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) ? Then the cleanup code in the finally clause can be removed.

Otherwise the changes look good to me!

cunningt: Thanks for the feedback - moved the finally code into a try-with-resources

thescouser89: LGTM! :)

thescouser89 (approved):

PR #922, opened 2022-03-28
Add DA report commands

jomrazek:
### Checklist:

### DIFF
--- a/da/src/main/java/org/jboss/bacon/da/DAReportsCli.java
+++ b/da/src/main/java/org/jboss/bacon/da/DAReportsCli.java
@@ -1,0 +1,223 @@
+package org.jboss.bacon.da;
+
+import lombok.extern.slf4j.Slf4j;
+import org.jboss.bacon.da.rest.endpoint.ReportsApi;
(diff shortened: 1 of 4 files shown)


da/src/main/java/org/jboss/bacon/da/DAReportsCli.java
```diff
@@ -0,0 +1,223 @@
...
+                ObjectHelper.print(false, result);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            return 0;
+        }
+    }
+
+    @CommandLine.Command(
+            name = "lookup-versions-npm",
```
jomrazek, line 183: @janinko we wanted to keep lookup for npm versions right? this one is not in new lookup endpoint

janinko: yup

thescouser89:
Our code is getting sexy with WLStripper! LGTM! THanks for the PR!

thescouser89 (approved):
LGTM

PR #926, opened 2022-04-11
[NCLSUP-624] make releaseStorage optional when unnecessary

michalovjan:
### Checklist:

### DIFF
--- a/integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/EmptyProjectBuildTest.java
+++ b/integration-tests/src/test/java/org/jboss/pnc/bacon/test/pig/EmptyProjectBuildTest.java
@@ -14,7 +14,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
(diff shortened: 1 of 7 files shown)


rnc:
Which commands is it now not needed for and which commands is it needed for? Can this also be clarified in the documentation to make it explicit ? Especially in relation to running the pig `run`, `release`, `cancel` (which CPaaS uses).

jomrazek:
it's needed for `configure`(`run`) if automatic milestone increment is enabled and for `release`. We need to improve documentation, releaseStorageUrl could be specified in build-config.yaml and in my opinion users should take care of it not cpaas in every command.

rnc:
> it's needed for `configure`(`run`) if automatic milestone increment is enabled and for `release`. We need to improve documentation, releaseStorageUrl could be specified in build-config.yaml and in my opinion users should take care of it not cpaas in every command.

At the moment CPaaS is having to pass it in all the time. So it sounds like we'll still need to keep doing that? 

Slightly off topic .... could it be specified in the config file thereby bypassing all of this confusion?

jomrazek:
You don't need to pass it for cancel after this change is tested and merged. Currently you don't need to pass it at all and it should be responsibility of users to use it in build-config.yaml if they need it for release command or they are using automatic milestone incrementation. I made it clear in our documentation that this is an option. [https://github.com/project-ncl/bacon/pull/929](url)

rnc:
> You don't need to pass it for cancel after this change is tested and merged. Currently you don't need to pass it at all and it should be responsibility of users to use it in build-config.yaml if they need it for release command or they are using automatic milestone incrementation. I made it clear in our documentation that this is an option. [#929](https://github.com/project-ncl/bacon/pull/929)

Thanks for the extra documentation. I don't think anyone was aware it was possible to use it in the configuration file and no file within the main build-configurations repo is currently using it in that way.

jomrazek (approved):
tested and everything seems to be working

PR #929, opened 2022-04-14
Make clear you can set releaseStorageUrl in build-config.yaml

jomrazek:
### Checklist:


guide/build-config.md
```diff
@@ -27,6 +27,7 @@ version: 1.0.0
 milestone: DR1
 brewTagPrefix: fb-1.0-pnc # Optional
 group: Test Product 1.0-all
+releaseStorageUrl: http://testurl.com/releaseStorage
```
rnc, line 30: `brewTagPrefix` above is marked as optional. Is this optional? What about any others?

guide/build-config.md
```diff
@@ -59,14 +60,19 @@ The product section describes the product that the builds belong. One example of
...
 
-The `--releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. This is url to be used in 'upload to candidates script' when running pig release command. Using '*' in your milestone you could also use this property for automatic incrementation of milestone value. For example 'ER*' as milestone will during pig run command check release storage and increment version to next available milestone version.
+The `releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. This is url to be used in 'upload to candidates script' when running pig release command. Using '*' in your milestone you could also use this property for automatic incrementation of milestone value. For example 'ER*' as milestone will during pig run command check release storage and increment version to next available milestone version. You can also set it with flag `--releaseStorageUrl=`
```
rnc, line 75: Minor fixes:
> This is the url to be used in the 'upload to candidates script' when running the `pig release` command

> Using `*` in the milestone means that this property may also be used for automatic increment of milestone value. For example using `ER*` as the milestone, during `pig run`, the command will check the configured release storage and increment the version to next available milestone version. It is possible to configure it either within the `build-config.yaml` or via the CLI using the flag `--releaseStorageUrl=<...>`.

Note the change in quote from ' to ` which solves an escaping issue on https://project-ncl.github.io/bacon/guide/build-config.html

Question - if both are configured which takes precedence?

guide/build-config.md
jomrazek: you don't ever need to set up brewTagPrefix that's why I think it is highlighted as optional, others are not optional in this sense and their description when they are used and user needs to set them is below.

guide/build-config.md
jomrazek: it's needed for `pig release` to use url in candidates upload script and then in `pig run` specifically `pig configure` if `*` is used in milestone property to check last available version and create new milestone. If you have specific milestone set e.g. `ER3` it's not needed here.

guide/build-config.md
rnc: So are you saying its now mandatory to provide `releaseStorageUrl` in the yaml ?

guide/build-config.md
```diff
@@ -59,14 +60,19 @@ The product section describes the product that the builds belong. One example of
...
 
-The `--releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. This is url to be used in 'upload to candidates script' when running pig release command. Using '*' in your milestone you could also use this property for automatic incrementation of milestone value. For example 'ER*' as milestone will during pig run command check release storage and increment version to next available milestone version.
+The `releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. This is the url to be used in the 'upload to candidates script' when running the `pig release` command. Using `*` in the milestone means that this property may also be used for automatic increment of milestone value. For example `ER*` as milestone, during `pig run`, the command will check the configured release storage and increment the version to next available milestone version. It is possible to configure it either within the `build-config.yaml` or via the CLI using the flag `--releaseStorageUrl=<...>`. Value in build-config.yaml takes precedence.
```
rnc, line 75:
```suggestion
The `releaseStorageUrl` is location of the release storage, typically on rcm-guest staging. This is the url to be used in the 'upload to candidates script' when running the `pig release` command. Using `*` in the milestone means that this property may also be used for automatic increment of milestone value. For example `ER*` as milestone, during `pig run`, the command will check the configured release storage and increment the version to next available milestone version. It is possible to configure it either within the `build-config.yaml` or via the CLI using the flag `--releaseStorageUrl=<...>` with the former taking precedence.
```

That is slightly surprising - normally CLI flags override pre-configured values?

jomrazek: We can actually change this in Jan's PR but at the moment it's like this

guide/build-config.md
jomrazek: Yes, it's needed if you want to run `pig release` or `pig configure/run` when using automatic milestone increment meaning you use `*` in milestone

rnc (approved):

PR #934, opened 2022-05-12
[NCL-7040] Add option to set dry-run build option from build-config.yaml

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
@@ -32,6 +32,7 @@ import org.jboss.pnc.bacon.pig.impl.out.PigReleaseOutput;
 import org.jboss.pnc.bacon.pig.impl.out.PigRunOutput;
 import org.jboss.pnc.bacon.pig.impl.pnc.ImportResult;
 import org.jboss.pnc.bacon.pig.impl.repo.RepositoryData;
+import org.jboss.pnc.bacon.pig.impl.utils.AlignmentType;
 import org.jboss.pnc.bacon.pig.impl.utils.FileDownloadUtils;
 import org.jboss.pnc.bacon.pnc.common.ParameterChecker;
 import org.jboss.pnc.enums.RebuildMode;
@@ -100,14 +101,13 @@ public class Pig {
         @Option(
                 names = { "-t", "--tempBuild" },
                 defaultValue = "false",
-                description = "If specified, PNC will perform temporary builds. Akin to dry-run but with temporary alignment preference.")
+                description = "If specified, PNC will perform temporary builds.")
         boolean tempBuild;
 
         @Option(
-                names = "--dry-run",
-                defaultValue = "false",
-                description = "If specified, PNC will perform a dry run. Akin to tempBuild but with persistent alignment preference.")
-        boolean dryRun;
+                names = { "--tempBuildAlignment" },
+                description = "Alignment preference for temporary build. Options: (default)TEMPORARY, PERSISTENT")
+        AlignmentType tempAlign = AlignmentType.TEMPORARY;
 
         @Option(
                 names = "--releaseStorageUrl",
@@ -259,7 +259,7 @@ public class Pig {
                     removeGeneratedM2Dups,
                     repoZipPath,
                     tempBuild,
-                    dryRun,
+                    tempAlign,
                     tempBuildTS,
                     RebuildMode.valueOf(rebuildMode),
                     skipBranchCheck,
@@ -338,7 +338,7 @@ public class Pig {
 
             ParameterChecker.checkRebuildModeOption(rebuildMode);
             GroupBuildInfo groupBuildInfo = PigFacade
-                    .build(tempBuild, tempBuildTS, RebuildMode.valueOf(rebuildMode), wait, dryRun);
+                    .build(tempBuild, tempBuildTS, RebuildMode.valueOf(rebuildMode), wait, tempAlign);
             if (wait) {
                 PigContext context = PigContext.get();
                 context.setBuilds(groupBuildInfo.getBuilds());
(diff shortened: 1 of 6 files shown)


rnc:
Why not allow `temporaryBuildAlignmentPreference=PERSISTENT|TEMPORARY` as per the ticket ? Would there be a matching documentation update?

jomrazek:
There is PR for bacon documentation, https://github.com/project-ncl/bacon/pull/938 , I will also work on separate documentation of build/dependencies behavior with dry-run option.

jomrazek:
> Why not allow `temporaryBuildAlignmentPreference=PERSISTENT|TEMPORARY` as per the ticket ? Would there be a matching documentation update?

I'm missing context or benefit of options PERSISTENT or TEMPORARY, I believe cpaas takes care of running build as temporary or persistent, this option just specify if it should be run as temporary dry-run or not, so it's simple to add property name and then just switch between true or false instead of thinking what is the other option to temporary etc.

rnc:
> > Why not allow `temporaryBuildAlignmentPreference=PERSISTENT|TEMPORARY` as per the ticket ? Would there be a matching documentation update?
> 
> I'm missing context or benefit of options PERSISTENT or TEMPORARY, I believe cpaas takes care of running build as temporary or persistent, this option just specify if it should be run as temporary dry-run or not, so it's simple to add property name and then just switch between true or false instead of thinking what is the other option to temporary etc.

I believe my suggestion is more explicit and allows the Bacon configuration to override the CLI option for all temporary use-cases. This is outside of CPaaS. However obviously its just my opinion that I think its clearer :-)

thescouser89:
I'll unfortunately add noise here. From what I see, if either `--dry-run` option or:
```yaml
    tempBuildPersistentAlign: true
```

is specified in the build-config.yaml, a temporary group build will happen with persistent alignment preference (no matter if the tempBuild flag is set or not in the bacon flags).

With Nick's suggestion, the setting will be set in the yaml and only activated when temporary builds happen (with flag --tempBuild).

I think I like Nick's suggestion more to be honest. It mimics what we use in the PNC UI also (persistent temp alignment option only appearing when temp build is selected).

rnc:
> I'll unfortunately add noise here. From what I see, if either `--dry-run` option or:
> 
> ```yaml
>     tempBuildPersistentAlign: true
> ```
> 
> is specified in the build-config.yaml, a temporary group build will happen with persistent alignment preference (no matter if the tempBuild flag is set or not in the bacon flags).
> 
> With Nick's suggestion, the setting will be set in the yaml and only activated when temporary builds happen (with flag --tempBuild).
> 
> I think I like Nick's suggestion more to be honest. It mimics what we use in the PNC UI also (persistent temp alignment option only appearing when temp build is selected).

To further clarify as I feel we have some misunderstanding - 

If the user had specified `--tempBuild` **and** `--dry-run` **but** had put  `temporaryBuildAlignmentPreference=TEMPORARY ` then the dry-run flag would be ignored.
If the user had specified `--tempBuild` **and** `--dry-run` **but** had put  `temporaryBuildAlignmentPreference=PERSISTENT ` then the dry-run flag is equivalent to the yaml.
If the user had specified `--tempBuild` **but** had put  `temporaryBuildAlignmentPreference=PERSISTENT ` temporary build with persistent alignment will be done
If the user had specified `--tempBuild` **and** `--dry-run` then a temporary build with persistent alignment will be done.

Perhaps we should all sync to discuss?

rnc:
Note also I entered NCL-7093 which affects this but also might make any implementation simpler.

pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
```diff
@@ -100,14 +100,14 @@ private Pig() {
...
 
         @Option(
-                names = "--dry-run",
-                defaultValue = "false",
-                description = "If specified, PNC will perform a dry run. Akin to tempBuild but with persistent alignment preference.")
-        boolean dryRun;
+                names = { "--tempBuildAlignment" },
+                defaultValue = "TEMPORARY",
+                description = "Alignment preference for temporary build. Options: (default)TEMPORARY, PERSISTENT")
+        String tempAlign;
```
rnc, line 110: I think this should be an enum I.e. https://picocli.info/#_enum_types
This would also simplify some of the following code and provide better error checking

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/pnc/PncBuilder.java
```diff
@@ -101,7 +101,12 @@ private GroupBuild run(
         if (dryRun) {
             buildParams.setTemporaryBuild(true);
             buildParams.setAlignmentPreference(AlignmentPreference.PREFER_PERSISTENT);
-        }
+            log.info("Build running as: dry-run temporary build");
```
rnc, line 104: I wonder if this should be explicit about the alignment preference?

rnc: I.e. "build running as a report build with persistent alignment preference"

pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
jomrazek: changed :+1:

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/config/PigConfiguration.java
```diff
@@ -79,6 +79,8 @@ public class PigConfiguration implements Validate {
     private Map<String, Map<String, ?>> addons = new HashMap<>();
     private String releaseStorageUrl;
 
+    private String temporaryBuildAlignmentPreference;
```
rnc, line 82: I presume this denotes the yaml config value - what happens if they set it to something other than temporary or persistent?

rnc (approved):
Looks good, thanks. Potentially there should be more unit tests for this change (as I'm on pto I'll leave that to you and @thescouser89  to discuss) but overall I think it's a good change

jomrazek:
@rnc great, do you need something else in the release? I would like to do it today if all fixes and changes you need are already included.

PR #935, opened 2022-05-12
Remove maven-embedder to fix dependency issue for licenses generator

jomrazek:
### Checklist:
licenses-generator uses org.jenkins-ci.lib:lib-jenkins-maven-embedder dependency which I would call obsolete, if there is no other suggestion how to solve this we need to dedicate time to upgrade licenses generator to use maven-embedder from apache or not upgrade above mvn 3.6.x as there is breaking change in ArtifactRepository (I believe)


thescouser89 (approved):

thescouser89:
It's basically reverting the changes that broken the license generator in the first place. LGTM!

PR #940, opened 2022-05-17
Add project logs

cunningt:
Adding project logs if 

printProjectLogs: true 

is set within the configuration.    Will add a PR to document it.


cunningt:
PR for the documentation of this change

rnc:
@cunningt I think the formatting for the PR is wrong ; I believe its possible to reformat but @thescouser89 or @jomrazek  would be able to clarify the command to run.

@thescouser89 In the build log, where it says `echo 'There are files with wrong formating. Exiting with FAIL` ; could that print the command the user's should run to correct their formatting?

jomrazek:
just `mvn clean install`

thescouser89:
Thanks @rnc for the suggestion! I have adjusted the Jenkins job to provide the instructions to fix it.

thescouser89:
Thanks @cunningt for the PR and documentation PR! :heart:

PR #957, opened 2022-06-27
[NCL-7146] Allow lookup of DA alignment versions via a filename

thescouser89:
The Quarkus team would like to automate their productization workflow by
generating a list of dependencies that need to be productized, and
lookup via `bacon da lookup maven-latest` whether those dependencies
already have a productized version or not.

It'll make their life much easier if they can specify the dependencies
(in GAV format) via a file (one dependency per line).

### Checklist:


dwalluck:
s/Aloow/Allow/

aloubyansky (approved):
Looks good @thescouser89, thanks!

PR #965, opened 2022-08-02
added --skipJsonOutput which if set turns off all the JSON output at …

paulgallagher75:
…the end of the PiG run, default is still how it was


rnc:
@paulgallagher75 There is already a `--jsonOutput` that is disabled by default - if that doesn't work doesn't that imply there is a bug there versus adding a new CLI option (cc @jomrazek )

PR #982, opened 2022-09-29
[NCL-7283] Make bacon pig run wait and not run multiple builds of sam…

jomrazek:
…e configuration in parallel

This is solving part of the problem, it's still possible for user to run multiple builds in parallel by using *pig run* if builds are started before first one enters building phase in pnc. I can't move check for running builds from start of the process to start of the building phase because we don't want to allow configuration change in pnc either. Solution might be to add table to pnc where bacon would send information which GroupConfig it will start build for and block it by reading this table.


rnc:
@jomrazek Honestly can't remember asking about this :-) I'm guessing this is to try and prevent clashes from CPaaS ? Your solution explanation looks good.

jomrazek:
@rnc you didn't :) just wanted your opinion and let you know it's happening. And yes, that is the reason.

thescouser89 (approved):
LGTM!

@jomrazek  I'll let you merge! Not sure if you want it to be part of main now or wait till 2.4.0 release. Up to you!

PR #986, opened 2022-10-14
Add a new CamelRuntimeDependenciesToAlignTree which can determine fir…

cunningt:
Add a new CamelRuntimeDependenciesToAlignTree which can determine first-level dependencies and filter out non-supported artifacts.     At the moment it is camel specific and filters artifacts on specific requirements that the Fuse teams have, but I'd like to parameterize it more in the future so that it could be used by other teams.

Added unit tests for new classes.

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/AddOnFactory.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/addons/AddOnFactory.java
@@ -17,6 +17,7 @@
  */
 package org.jboss.pnc.bacon.pig.impl.addons;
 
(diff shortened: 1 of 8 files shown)


thescouser89:
Hi @cunningt ! Could you please just run a `mvn clean install -DskipTests=true`, then rebase your commit with the changes?
The PR failure is only because of code formatting issues, which the `mvn clean install` will fix for you. Thanks!

thescouser89:
The camel.TreeParserTest.testLargerScaleTreeParser test is failing :(

```
Expected to find 3922 dependencies, # of dependencies was 3854 ==> expected: <true> but was: <false>
```

cunningt:
Fixed test; refreshed to main

PR #988, opened 2022-10-18
Check for product version change and update group configuration accor…

jomrazek:
…dingly

### Checklist:


thescouser89 (approved):
LGTM!

Is this handling some kind of edgecase we hit in Prod where the production version changed for the same build configuration name?

jomrazek:
@thescouser89 from what I saw, most of the teams are using version property in group name so they never hit it and that's why we or at least I never noticed we're not updating existing group with this info. Petr has I guess some test case for deliverable analyzer where he needs this as he complained about it and as it is possible through UI I implemented it also into bacon if some team wants to use single group. If there is some reason why it wasn't implemented in the past we can close this too.

thescouser89:
Sounds good to me!

PR #998, opened 2022-11-16
NCL-7430 Add OpenTelemetry instrumentation

rnc:
### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/bacon/common/OTELHelper.java
+++ b/common/src/main/java/org/jboss/pnc/bacon/common/OTELHelper.java
@@ -1,0 +1,71 @@
+package org.jboss.pnc.bacon.common;
+
+import com.redhat.resilience.otel.internal.EnvarExtractingPropagator;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -81,6 +81,9 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
 
+        <otel.alpha.version>1.19.0-alpha</otel.alpha.version>
```
vibe13, line 84: 1.19.2 is out

common/src/main/java/org/jboss/pnc/bacon/common/OTELHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    public void stopOtel() {
+        if (root != null && processor != null) {
+            log.debug("Finishing OTEL instrumentation for {}", root);
+            root.end();
+            processor.close();
+        }
+    }
+
+    public String generateTraceParent(String trace, String span) {
+        return "00" + '-' + trace + '-' + span + "-01";
```
vibe13, line 77: Ok this last "-01" might vary (depends mainly if the spans are sampled or not). I don't have a logic for that yet, but we might need to change that in the future, for now we can consider it ok!

vibe13:
LGTM for what I can tell!

vibe13 (changes requested):

pom.xml
rnc: Thanks! Actually I don't think I need that variable ( used it when testing the autoconfigurespi)

common/src/main/java/org/jboss/pnc/bacon/common/OTELHelper.java
rnc: Sure that makes sense.

vibe13 (approved):

PR #1001, opened 2022-11-21
Include artifacts with JSON extensions into the Maven repo ZIP in the RESOLVE_ONLY strategy impl

aloubyansky:
It looks like (some of the) other strategies do support it but the RESOLVE_ONLY one for some reason wasn't tested for it.

### Checklist:


jomrazek:
I'm not against this but I would like to have opinion of other productization teams. @paulgallagher75 maybe somebody from your team or other product teams you could recommend

aloubyansky:
@paulgallagher75 we discussed this issue on a call last week. It's about collecting productized JSON artifacts present in BOMs to include in the Maven repo ZIP.

The reason it wasn't collecting JSON artifacts for us was that it **assumed** that if an **artifactId** contained `bom` it's a POM artifact. This is a trap for the future, btw. I am not sure why this was implemented this way. For now I simply made sure it'll work for our current use-cases.

jomrazek (approved):

PR #1009, opened 2022-12-14
Set IncludeNotFound to false as not found artifacts are ignored anyway

janinko:
### Checklist:


dwalluck (approved):
I am not sure if this makes any difference at all, but it shouldn't hurt.

It just controls whether build id 0 (not found builds) is included in the returned list, but as you state, this is filtered when generating the NVR report anyway (https://github.com/project-ncl/build-finder/blob/d9dd5a2b777c33ca5d52ae38fa0171f4c2e1a0cf/report/src/main/java/org/jboss/pnc/build/finder/report/NVRReport.java#L54).

janinko:
retest this please

PR #1015, opened 2023-01-03
Adding changes for build config generator

harshmadhani:
### Checklist:

### DIFF
--- a/temp-name/src/main/java/org/jboss/bacon/tempname/impl/generator/BuildConfigGenerator.java
+++ b/temp-name/src/main/java/org/jboss/bacon/tempname/impl/generator/BuildConfigGenerator.java
@@ -2,10 +2,18 @@ package org.jboss.bacon.tempname.impl.generator;
 
 import org.jboss.bacon.tempname.impl.config.BuildConfigGeneratorConfig;
 import org.jboss.bacon.tempname.impl.dependencies.DependencyResult;
+import org.jboss.bacon.tempname.impl.dependencies.Project;
+import org.jboss.bacon.tempname.impl.projectfinder.FoundProject;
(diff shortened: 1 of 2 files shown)


temp-name/src/main/java/org/jboss/bacon/tempname/impl/generator/BuildConfigGenerator.java
```diff
@@ -2,17 +2,43 @@
...
+        Set<Project> projects = dependencies.getTopLevelProjects();
+        Set<GAV> gavsToSearch = new HashSet<>();
+        Iterator<Project> setIterator = projects.iterator();
+        while (!projects.isEmpty() && setIterator.hasNext()) {
+            gavsToSearch.addAll(setIterator.next().getGavs());
+        }
+
+        Set<FoundProject> foundProjectSet = foundProjects.getFoundProjects();
+        Iterator<FoundProject> foundProjectIterator = foundProjectSet.iterator();
+        BuildConfig tempBC;
```
janinko, line 35: You can move the declaration inside the cycle if it is not used outside the cycle.

temp-name/src/main/java/org/jboss/bacon/tempname/impl/generator/BuildConfigGenerator.java
```diff
@@ -2,17 +2,43 @@
...
+        while (!foundProjectSet.isEmpty() && foundProjectIterator.hasNext()) {
+            tempBC = BuildConfigMapping.toBuildConfig(
```
janinko, line 37: Also I think you would want to call the method with takes the BC revison too, so that we have the revison that is closest to the version we want to build.

PR #1023, opened 2023-01-19
feat(autobuild): use the AUTO build category for all autobuild create…

goldmann:
…d BuildConfigurations

### Checklist:


janinko:
Do we want to use the AUTO category for Phase 2? Where there might not be the qualities like fully formed depenendency tree?

janinko:
Also, is the category already available in prod?

goldmann:
Good points.

> Do we want to use the AUTO category for Phase 2? Where there might not be the qualities like fully formed depenendency tree?

Probably we don't want to have this. This would potentially mean we will populate the AUTO build category with some garbage that would be hard to track.

> Also, is the category already available in prod?

No, it's only on devel. Not even in stage.

PR #1026, opened 2023-01-24
Add experimental Autobuilder implementation

janinko:
### Checklist:


jomrazek (commented):
looks ok from bacon perspective but please make that test pass or if it's not possible and should be just some local test disable it for default build.

PR #1042, opened 2023-03-08
Upgrade to Domino 0.0.79 and Quarkus 2.16.4.Final

aloubyansky:
* 0.0.79 includes a critical fix when sorting project releases for the build. It is now taking into account *only* the direct dependencies of the artifacts as their appear in their POMs, instead of also those that appear as their direct dependencies in the dependency trees generated for the root project artifacts;
* Quarkus resolver 2.16.4.Final is what Domino is currently using.

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
 
     <properties>
         <compiler.plugin.version>3.8.1</compiler.plugin.version>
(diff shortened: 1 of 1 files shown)


aloubyansky:
FYI @janinko

PR #1057, opened 2023-04-12
[AB] Upgrade deprecated environments based on the provided replacement

janinko:
### Checklist:

### DIFF
--- a/experimental/src/main/java/org/jboss/bacon/experimental/impl/generator/BuildConfigGenerator.java
+++ b/experimental/src/main/java/org/jboss/bacon/experimental/impl/generator/BuildConfigGenerator.java
@@ -214,7 +214,8 @@ public class BuildConfigGenerator {
     }
 
     public BuildConfig copyExisting(BuildConfiguration bc, BuildConfigurationRevision bcr, String name) {
-        boolean useEnvironmentName = shouldUseEnvironmentName(name, bcr.getEnvironment());
+        Environment env = environments.resolve(bcr.getEnvironment());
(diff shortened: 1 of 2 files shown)


jomrazek (commented):
Are you aware this could potentially break builds?

janinko:
It should not break builds as the upgrade path should be veted by PNC admins to be safe.  But yes.

experimental/src/main/java/org/jboss/bacon/experimental/impl/generator/BuildConfigGenerator.java
```diff
@@ -233,8 +235,26 @@ public BuildConfig copyExisting(BuildConfiguration bc, BuildConfigurationRevisio
         return buildConfig;
     }
 
+    private static void setEnvironment(BuildConfig buildConfig, Environment env, boolean useName) {
+        if (useName) {
+            if (!env.getName().equals(buildConfig.getEnvironmentName())) {
+                log.debug(
```
jomrazek, line 241: is it worth to change it into info level?

jomrazek (approved):

experimental/src/main/java/org/jboss/bacon/experimental/impl/generator/BuildConfigGenerator.java
janinko: well, Kuba said that it should not bother users, but sure

jomrazek: well I would say it's informing not bothering and if they don't want to be bothered they can update their environment. Thank you

PR #1072, opened 2023-06-29
[NCL-7258] Don't modify the builds map for source generation

thescouser89:
Depending on the source generation, we may download sources for the builds done as part of the group build in PiG + other sources as specified by the user.

Unfortunately we were modifying the global `builds` map that should only contain the builds done in a group build. This in turn, confused CPaaS into what builds were actually done.

This commit introduces a shallow copy of the builds map so that we modify this instead to add any additional builds to to download sources instead of modifying the global one.

### Checklist:


jomrazek (approved):
Thank you Dustin!

PR #1076, opened 2023-07-12
[NCL-7989] allow to exclude specific builds in sources generation

rgdoliveira:
Jira: https://issues.redhat.com/browse/NCL-7989

Related PRs:
- https://github.com/project-ncl/bacon/pull/1076
- https://github.com/project-ncl/bacon/pull/1077

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/sources/SourcesGenerator.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/sources/SourcesGenerator.java
@@ -77,11 +77,6 @@ public class SourcesGenerator {
             return;
         }
 
-        if (!sourcesGenerationData.getExcludeSourceBuilds().isEmpty()) {
-            log.info("Removing builds from sources generation " + sourcesGenerationData.getExcludeSourceBuilds());
-            additionalBuildsForSources.keySet().removeAll(sourcesGenerationData.getExcludeSourceBuilds());
-        }
-
         File workDir = FileUtils.mkTempDir("sources");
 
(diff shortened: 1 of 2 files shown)


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/sources/SourcesGenerator.java
```diff
@@ -77,6 +77,11 @@ public void generateSources(Map<String, PncBuild> builds, RepositoryData repo) {
             return;
         }
 
+        if (!sourcesGenerationData.getExcludeSourceBuilds().isEmpty()) {
+            log.info("Removing builds from sources generation " + sourcesGenerationData.getExcludeSourceBuilds());
+            additionalBuildsForSources.keySet().removeAll(sourcesGenerationData.getExcludeSourceBuilds());
+        }
+
```
jomrazek, line 84: Looks good although I would ask if we don't want to exclude artifacts after all options of their addition is done = exclude them after all strategies to generate them are done.

rgdoliveira: @jomrazek that's a good point, I guess it is better to exclude after all strategies are generated (as you suggested) so we are sure that what is added to the `excludeSourceBuilds` will be really removed no matter what.

I pushed a new commit with your suggestion.

rgdoliveira: @jomrazek and the PR updating the documentation is https://github.com/project-ncl/bacon/pull/1077

jomrazek (approved):

PR #1085, opened 2023-09-07
[NCL-8165] Allow to suppply license exception jsons as parameter

jomrazek:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
@@ -243,6 +243,16 @@ public class Pig {
                 description = "If set, any missing source jars will throw an error")
         private boolean strictDownloadSource;
 
+        @Option(
+                names = "--licenseExceptionsPath",
+                description = "Path to file with exceptions for license generation. Usually rh-license-exceptions.json.")
+        private String licenseExceptionsPath;
+
+        @Option(
+                names = "--licenseNamesPath",
+                description = "Path to file with license names for license generation. Usually rh-license-names.json.")
+        private String licenseNamesPath;
+
         /**
          * addOns to skip. Specified in CLI by using the flag multiple times: --skipAddon=a --skipAddon=b
          *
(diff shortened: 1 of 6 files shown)


jomrazek:
Hey @saumya1singh , would you mind testing this PR for your license generations?

you can either fetch the PR and build it while removing/moving pig/src/main/resources/rh-license-exceptions.json and rh-license-names.json or use jar file here:

https://drive.google.com/file/d/1YPcOgg4E6NJnr55qmQ6OSBhDn0zP2pXl/view?usp=sharing

from there 
`java -jar *path*/bacon-license-test.jar pig licenses *your usual commands* --licenseExceptionsPath=*path*/rh-license-exceptions.json --licenseNamesPath=*path*/rh-license-names.json`

(you will probably need to run `pig repo` before generating licenses)

thescouser89:
Does the flag replace the one built into bacon or will be considered on top of the one included in bacon?

saumya1singh:
Hi @jomrazek ,
Thank you for the PR, I appreciate your efforts on this. 👏🏻 

My only concern is we (Quarkus Prod) don't want to do license separately, it create issues for us. 

I will try to add `--licenseExceptionsPath=*path*/rh-license-exceptions.json --licenseNamesPath=*path*/rh-license-names.json`  params in the jenkins script and see how it goes. 😄

saumya1singh:
Hi @jomrazek , 

In the path can we give urls also? That will be very handy for me, I will be able to add that in the Jenkins Script directly. Something like 

`--licenseExceptionsPath=https://github.com/project-ncl/bacon/blob/main/pig/src/main/resources/rh-license-exceptions.json --licenseNamesPath=https://github.com/project-ncl/bacon/blob/main/pig/src/main/resources/rh-license-names.json`

jomrazek:
> Does the flag replace the one built into bacon or will be considered on top of the one included in bacon?

yes, it should take preference over built in files

jomrazek:
> Hi @jomrazek ,
> 
> In the path can we give urls also? That will be very handy for me, I will be able to add that in the Jenkins Script directly. Something like
> 
> `--licenseExceptionsPath=https://github.com/project-ncl/bacon/blob/main/pig/src/main/resources/rh-license-exceptions.json --licenseNamesPath=https://github.com/project-ncl/bacon/blob/main/pig/src/main/resources/rh-license-names.json`

Hi, currently no urls, just local files. Also this should not be a way to separate license effords of different teams but to provide you a quicker way of going through your productization process without need of commiting and waiting for snapshot build every time you fix one license. At the end when you fix all licenses for your version/build you're always encouraged to push your updated files into bacon repo.

thescouser89 (approved):

PR #1087, opened 2023-09-13
Clean files created during testruns

jomrazek:
There is a lot of noise after running this test, this cleanup will remove files created by the test.
Screenshot from 2023-09-13 14-47-05


thescouser89 (approved):
LGTM!

PR #1082, opened 2023-08-30
Disallowing builds with specific suffix to the

harshmadhani:
### Checklist:

* [ ] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) for your change if user-facing?
* [ ] Have you added unit tests for your change?

PR #1097, opened 2023-11-02
Make source download name more unique

jomrazek:
this was failing for existing file when user used mutliple versions built by same build configuration = oss-parent-50-redhat-9 and oss-parent-50-redhat-7


zhfeng:
It might be related to
- https://github.com/project-ncl/bacon/pull/1082

thescouser89 (approved):

PR #1105, opened 2023-12-22
BOM Maven repository generation strategy that's based on parallel dependency resolution

aloubyansky:
This change introduces a new Maven repository resolution strategy called `BOM` for BOM-based products.
Essentially, the BOM strategy is an enhancement of the `RESOLVE_ONLY` strategy enabling parallel dependency resolution and repository content processing, while re-using the same configuration of the `RESOLVE_ONLY` strategy.

BOM strategy allows to save ~1h on Maven repo ZIP generation for RHBQ 3.2.9 release with the following command
```
java -jar /home/aloubyansky/git/bacon/cli/target/bacon.jar pig run --clean --skipBuilds --skipBranchCheck --skipPncUpdate --skipLicenses --skipSources --releaseStorageUrl https://download.eng.bos.redhat.com/rcm-guest/staging/quarkus /home/aloubyansky/playground/bacon/ --targetPath /home/aloubyansky/playground/bacon/deliverables-bom
```

Reducing the total time to ~11 min from 70+ min (the actual time will depend on the environment, network and state of remote services). To reproduce the result locally, replace `RESOLVE_ONLY` at [1] with `BOM`.

[1] https://gitlab.cee.redhat.com/middleware/build-configurations/-/blob/master/Quarkus/quarkus-platform/3.2/build-config.yaml#L40

### Checklist:


jomrazek (commented):
Honestly this is quite a big change to existing code for me to be OK with, mainly in resolveAndRepackage method and similar cases when method is used by basically all other strategies and could potentially break generation for many teams. I don't really have capacity to test every corner case teams discovered and fixed or hacked to work for them in current workflow in my brain :) Given my experience with similar improvements and then us being responsible of fixing bugs in production I would be more in favor of creating flag and separating new improved generation approach  into own separated methods (despite temporary code duplicities) for teams to test faster generation with but also with clear way of falling back to old approach if something is not working. Unfortunately nobody will test this in snapshot version but then it will become blocker that needs immediate fix if something is broken in released version. Just my opinion

aloubyansky:
@jomrazek I share your concerns which is why I tried to follow your suggestion of introducing alternative method implementations instead of re-implementing existing strategies and methods. E.g. checksum generation.

But let's please be more specific about what should remain unchanged and why.
If you see a method that is used by repo generation strategies other than `RESOLVE_ONLY` and the new `BOM` one that has changed in a way that raises a concern please point it out.

AFAICS, `resolveAndRepackage` is used only by `RESOLVE_ONLY` strategy which, AFAIR, was implemented for Quarkus, is also BOM-based, and even had (unfortunate) Quarkus-related terms in variable names. The only change in this method is related to collection of artifacts from a BOM (applying filtering based on `redhat` version qualifier and include/exclude filters) whose dependencies should be resolved.

`parseExtensionsArtifactList` is also Quarkus-specific (just by the name of it) and the only change there is closing the input stream.

`collectRedHatVersions()` was refactored to `ensureVersionsSet` to
1) reflect better what it is actually doing
2) return a consumable list of artifacts with versions set instead of a map of artifact key to version

Would you like to chat about these changes? Or perhaps have more specific requests to adjust it?
When it comes to fixing regressions in this specific code, if any, you could ping me directly.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepoManager.java
```diff
@@ -520,65 +531,46 @@ private void resolveAndRepackage(
...
-                    .getOrDefault(extensionRtArtifact, extensionRtArtifact.getVersion());
-            if (extensionRtArtifact.getArtifactId().contains("plugin")) {
-                Artifact pluginArtifact = new DefaultArtifact(
-                        extensionRtArtifact.getGroupId(),
-                        extensionRtArtifact.getArtifactId(),
-                        extensionRtArtifact.getExtension(),
-                        aptVersion);
-                log.debug("Resolving dependencies of plugin {}", pluginArtifact);
-                mvnResolver.resolvePluginDependencies(pluginArtifact);
+            if (artifact.getArtifactId().contains("maven-plugin")) {
```
jomrazek, line 558: I don't think this is matching same artifacts as before. Some plugin artifacts doesn't have "maven-plugin" in name, is it intentional?

jomrazek (commented):
You're right about it, we have different repackage method used in multiple strategies, went through whole class and refactored code seems to do equivalent thing apart that one comment. Other than that looks like is should not break something existing.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepoManager.java
aloubyansky: Yes, it is intentional. I found that some artifacts had `plugin` in the artifactId name that weren't a Maven plugin and their dependencies weren't properly resolved.

aloubyansky: Either way, this check is a hack, unfortunately, and needs to be handled in some better way.
A proper way would be to have a dedicated config option for artifacts that are Maven plugins or this code has to resolve the artifact and check whether it contains Maven plugin metadata, which will be less efficient but more correct.

aloubyansky: As the rest of it, this is a Quarkus-specific part. The reason Maven plugin dependencies are resolved differently from the rest of the artifacts is that, the BOM configured as the source of dependencies to resolve should not be enforced on Maven plugin artifacts, as it wouldn't happen in customer applications. But for the rest of the artifacts the BOM has to be enforced while resolving dependencies.

jomrazek (approved):
Given the changes only to single strategy we cant try to merge this, the plugin stuff can still in my opinion unexpectedly change results for other teams using this strategy but that's more decision of teams using it to decide what is correct for them and not mine.

aloubyansky:
I don't think there are other teams using these strategies, tbh. For Quarkus this `plugin` adjustment is necessary either way.

janinko (approved):

aloubyansky:
Is it good to merge?

thescouser89 (approved):

thescouser89:
I am not able to judge if this will break things or not. I'm too unfamiliar with this codebase. Sorry! I have approved this PR though since I also cannot prove this won't break things.

@aloubyansky Could you also please create a PR on our gh-pages branch to mention the new alternative to RESOLVE_ONLY for our docs? Thanks!

jomrazek:
PR in fact broke license generation, probably due to dep updates(?) snowdrop is fairly sensitive. I reverted it in main to unblock user who use snapshot because of license exceptions. Did you try full run of generations when testing or just repo?

`[INFO ] - Generating licenses
[DEBUG] - Using manager EnhancedLocalRepositoryManager with priority 10.0 for /home/jmrazek/.m2/repository
Exception in thread "main" java.lang.NoSuchMethodError: 'org.eclipse.aether.repository.RemoteRepository$Builder org.eclipse.aether.repository.RemoteRepository$Builder.setBlocked(boolean)'
	at org.apache.maven.RepositoryUtils.toRepo(RepositoryUtils.java:213)
	at org.apache.maven.bridge.MavenRepositorySystem.getProxy(MavenRepositorySystem.java:279)
	at org.apache.maven.bridge.MavenRepositorySystem.injectProxy(MavenRepositorySystem.java:310)
	at org.apache.maven.internal.aether.DefaultRepositorySystemSessionFactory.newRepositorySession(DefaultRepositorySystemSessionFactory.java:234)
	at org.apache.maven.DefaultMaven.newRepositorySession(DefaultMaven.java:350)
	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:123)
	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:110)
	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:137)
	at me.snowdrop.licenses.maven.SnowdropMavenEmbedder.<init>(SnowdropMavenEmbedder.java:32)
	at me.snowdrop.licenses.maven.MavenEmbedderFactory.getSnowdropMavenEmbedder(MavenEmbedderFactory.java:30)
	at me.snowdrop.licenses.LicensesGenerator.<init>(LicensesGenerator.java:75)
	at org.jboss.pnc.bacon.pig.impl.license.LicenseGenerator.generateLicenses(LicenseGenerator.java:81)
	at org.jboss.pnc.bacon.pig.impl.license.LicenseGenerator.generateLicenses(LicenseGenerator.java:67)
	at org.jboss.pnc.bacon.pig.impl.license.LicenseManager.generate(LicenseManager.java:117)
	at org.jboss.pnc.bacon.pig.impl.license.LicenseManager.prepare(LicenseManager.java:83)
	at org.jboss.pnc.bacon.pig.PigFacade.generateLicenses(PigFacade.java:576)
	at org.jboss.pnc.bacon.pig.PigFacade.run(PigFacade.java:264)
	at org.jboss.pnc.bacon.pig.Pig$Run.doExecute(Pig.java:274)
	at org.jboss.pnc.bacon.pig.Pig$Run.doExecute(Pig.java:174)
	at org.jboss.pnc.bacon.pig.Pig$PigCommand.call(Pig.java:162)
	at org.jboss.pnc.bacon.pig.Pig$PigCommand.call(Pig.java:97)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2314)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
	at org.jboss.pnc.bacon.cli.App.executionStrategy(App.java:288)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at org.jboss.pnc.bacon.cli.App.run(App.java:213)
	at org.jboss.pnc.bacon.cli.App.main(App.java:88)
`

aloubyansky:
No, I was testing only the repo generation

aloubyansky:
Where is snowdrop code repository located?

aloubyansky:
https://github.com/snowdrop/licenses-generator
It looks like it's pretty much unmaintained and depends on Maven 3.5.4.
That'd be another tool to modernize.

aloubyansky:
@jomrazek please see https://github.com/project-ncl/bacon/pull/1108

PR #1116, opened 2024-03-05
Replace config variables used as auto-build configuration scalars

MikeEdgar:
This change allows for tagged entries in the experimental autobuild configuration file to reference either system properties or environment variables. The use case I have in mind is where a project POM is missing the `scm` tag section, which currently defaults to `master` for the generated build config. 

Using a variable in the autobuild input configuration would allow for the correct revision to be set from some value, e.g. the git tag or commit hash in a CI process.

### Checklist:


thescouser89:
@MikeEdgar Could you provide an example on how that would look please?
And how does it defer from: https://project-ncl.github.io/bacon/guide/build-config.html#usage-of-yaml-variables

thescouser89:
retest this please

thescouser89:
retest this please

MikeEdgar:
@thescouser89 if it doesn't seem generally useful, I am fine to close this one. We have a kind of workaround for it at the moment so it's not a blocker.

PR #1121, opened 2024-04-03
Fix to return the prop with exceptionsPath and namesPath

zhfeng:
### Checklist:


svkcemk:
@thescouser89 @jomrazek FYI

thescouser89:
Ah good catch!!!

PR #1141, opened 2024-04-19
Use final-log for PiG and align-logs

thescouser89:
This commit removes the use of Bifrost live-logs feature for:
- PiG (since the final logs will already be present and we should use it for speed and reducing load on bifrost db search)
- align-logs (since we don't have a follow mode and we probably will invoke it after the alignment is done)

### Checklist:


jomrazek (approved):

thescouser89:
Oh I might have to keep this open until it's close to final-log release after all.

thescouser89:
@jomrazek if 3.0 comes out while I'm out, could you please merge this? It's not a big deal though if it doesn't get merged for 3.0. It can be done later. Thanks!

jomrazek:
Will do, no worries :)

PR #1160, opened 2024-05-09
feat: implement a global excludes to exclude specific builds from...

jacobdotcosta:
…autobuilder

### Checklist:


aloubyansky:
@janinko fyi

AntStephenson:
Sorry, I know you haven't been at the meetings so you missed some context to this ticket, I've updated it and added some extra details on the requirements. But in short, this global list should be in GitLab somewhere, rather than being configured on the build config side

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResult.java
```diff
@@ -21,4 +23,15 @@ private void fillAllProject(Set<Project> allProjects, Project project) {
             project.getDependencies().forEach(p -> fillAllProject(allProjects, p));
         }
     }
+
+    /**
+     * Remove the provided GAV from the {@link #topLevelProjects} set.
+     *
+     * @param excludedGavs GAV to be excluded from the topLevelProjects.
+     */
+    public void removeProjects(String[] excludedGavs) {
```
janinko, line 32: why is there the method?

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResolver.java
```diff
@@ -55,6 +57,13 @@ public DependencyResolver(DependencyResolutionConfig dependencyResolutionConfig)
 
     private void setupConfig(ProjectDependencyConfig.Mutable dominoConfig) {
         config.getExcludeArtifacts().stream().map(GACTVParser::parse).forEach(dominoConfig::addExcludePattern);
+        DependencyExcluder dependencyExcluder = new DependencyExcluder(Config.instance().getActiveProfile().getDa());
+        final String[] excludedGavs = DependencyExcluder.getExcludedGavs(dependencyExcluder.fetchExclusionFile());
+        log.info("There are {} dependencies to be excluded", excludedGavs.length);
+        Arrays.stream(excludedGavs)
+                .iterator()
+                .forEachRemaining(excludedGav -> dominoConfig.addExcludePattern(GACTVParser.parse(excludedGav)));
```
janinko, line 65:
```suggestion
        Arrays.stream(excludedGavs)
                .map(GACTVParser::parse)
                .forEach(dominoConfig::addExcludePattern);
```

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResolver.java
```diff
@@ -55,6 +57,13 @@ public DependencyResolver(DependencyResolutionConfig dependencyResolutionConfig)
 
     private void setupConfig(ProjectDependencyConfig.Mutable dominoConfig) {
         config.getExcludeArtifacts().stream().map(GACTVParser::parse).forEach(dominoConfig::addExcludePattern);
+        DependencyExcluder dependencyExcluder = new DependencyExcluder(Config.instance().getActiveProfile().getDa());
```
janinko, line 60: I think that creating new config class for Autobuilder would be better, then reusing a Dependency Analyzer config.

Also I wonder, if this config may be optional, so people not using autobuilder don't need it, but require it for the autobuilder.

janinko (changes requested):

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResolver.java
jacobdotcosta: I'll create the autobuild class. I've written a specific test to make sure builds don't break if the configuration is missing.

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResult.java
jacobdotcosta: Fixed

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResolver.java
jacobdotcosta: Changed the code.

experimental/src/main/java/org/jboss/bacon/experimental/impl/dependencies/DependencyResolver.java
jacobdotcosta: Done. I've created a new config class for autobuild. It is optional and shouldn't fail if isn't informed. I've created a test case for this situation and it seems to work correctly.

PR #1167, opened 2024-05-16
Update source generation to use consistent format for directory structure

AntStephenson:
Update the source generation so directory naming is more consistent, because its derived from the build config there are inconsistencies in how people name them. For example, one directory is simply called `2.2.0` without any reference to the groupId or artifactId.

This PR updates the naming to be derived from the BREW_BUILD_NAME and BREW_BUILD_VERSION which to my knowledge should be available as its automatically set, and I have double checked numerous builds including some gradle ones and it was, but if they aren't set it will leave them as is.

### Checklist:


AntStephenson:
@janinko @thescouser89 would someone mind reviewing this if you have time please? 💪🏻

jomrazek (approved):

PR #1218, opened 2024-08-21
Fixed GAV(String artifactPath) impl to properly parse the classifier

aloubyansky:
This change also avoids generating stacktraces when a classifier is empty, which is more efficient.

### Checklist:


aloubyansky:
Ok, this fixes highlights another issue that needs to be fixed. I'll open another PR that will include more fixes.

PR #1219, opened 2024-08-21
Remove maven-metadata.xml from checksums and GAV(path) fixes

aloubyansky:
1. Remove `maven-metadata.xml` from the constant that represents checksum extensions `RepoDescriptor.CHECKSUM_EXTENSIONS` (while it's used by some methods to identify checksum files, in other places it appears to be misused to filter out non-artifact files)
2. Fixed artifact filtering in `RepoDescriptor.listGavs(File dir)`, it doesn't filter out `_remote.repositories` as a non-artifact file (since it relied on `RepoDescriptor.CHECKSUM_EXTENSIONS`) which results in parsing it as
```
System.out.println(new GAV("org/jboss/jboss-parent/38/_remote.repositories").toGapvc());
```
```
org.jboss:jboss-parent:repositories:38:_remote
```
and including it in the results

3. Added a method `RepoDescriptor.listArtifacts(Path m2RepoDirectory)` that allows collecting complete artifact coordinates (instead of only GAVs as `listGavs(File dir)` does) allowing `m2RepoDirectory` to be either a directory or a ZIP file

4. Fixed parsing of a classifier in `GAV` and avoid generating stacktraces when a classifier is empty, which makes it more efficient.

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepoDescriptor.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepoDescriptor.java
@@ -20,9 +20,13 @@ package org.jboss.pnc.bacon.pig.impl.repo;
 import org.jboss.pnc.bacon.pig.impl.utils.GAV;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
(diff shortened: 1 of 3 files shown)


thescouser89:
LGTM!

PR #1225, opened 2024-08-27
Fix NotYetAlignedFromDependencyTree

thescouser89:
This commit fixes the integration between getting logs from Bifrost (via database and not from the final-log feature) and the NotYetAlignedFromDependencyTree.

Previously, the addon 'NotYetAlignedFromDependencyTree' expected the line to analyze to begin with:
```
[INFO] +
```

But Bifrost adds an extra info for some reason. The logs then become:
```
[INFO] [INFO] +
```

which doesn't match the line we want. This commit loosens the requirement to just have the '[INFO] +' line present somewhere.

Unit tests are written to make sure this logic is codified properly also.

### Checklist:


jomrazek (approved):
:+1:

jomrazek:
just needs formatting :)

PR #1231, opened 2024-09-03
[NCL-7601] Translate artifact path to identifier instead of the identifier regex to path regex in RepoUtils#removeExcludedArtifacts

patrikk0123:
### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
@@ -44,6 +44,7 @@ import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


patrikk0123:
The original problem was with the translation of `excludeArtifacts` regex in Maven artifact format to regex in file path format, which was not fully equivalent.

For example, this regex `'.*:.*:.*:.*:(?!db-scripts).*'` => `.*/.*/.*/.*-.*-(?!db-scripts).*\..*$`
the nonpresence of `db-scripts` is checked at the wrong place because of the `.*` and multiple `-` that may be in the input string

Therefore, I did it the other way around by translating the artifact path to the identifier.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
```diff
@@ -344,41 +345,36 @@ public static void recursivelyDeleteEmptyFolder(File element) {
...
-    public static String convertMavenIdentifierToPathRegex(String identifier) {
-        String[] sections = identifier.split(":");
-        if (sections.length != 4 && sections.length != 5) {
-            throw new IllegalArgumentException(
-                    identifier
-                            + " is wrongly formatted! It should be <group-id>:<artifact-id>:<packaging>:<version>:[classifier]");
-        }
-        String fileSeparator = System.getProperty("file.separator");
+    public static String convertArtifactPathToIdentifier(String artifactPath) {
+        var fileSeparator = System.getProperty("file.separator");
```
jomrazek, line 357: Please change this to `File.separator` , we tend to use that when accessing the default separator on other places, also by doing so you can get rid of this variable and use it directly on next line.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
```diff
@@ -243,19 +244,19 @@ public static void removeCommunityArtifacts(File element) {
...
-                Pattern matchesRegex = Pattern.compile(excludedPath);
-                if (matchesRegex.matcher(element.getAbsolutePath()).find()) {
+            for (String exclusionRegex : excludedArtifacts) {
+                Pattern matchesRegex = Pattern.compile(exclusionRegex);
+                var artifactIdentifier = convertArtifactPathToIdentifier(element.getAbsolutePath());
+
+                if (matchesRegex.matcher(artifactIdentifier).find()) {
                     log.debug(
-                            "Removing path {} from the repository since it matches regex: {}",
+                            "Removing path {} from the repository since it the artifact {} matches regex: {}",
```
jomrazek, line 256: 'it' is obsolete now?

jomrazek (commented):
Not a fan of using `var` , small comments, otherwise looks good to me.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/repo/RepositoryUtils.java
patrikk0123: I had to do it like this: `Pattern.quote(File.separator)`
Because on Windows, file separator is a `\` which is not a valid regex

thescouser89:
LGTM! Converting `removeExcludedArtifacts` to an identifier to apply the regex. + unit tests for safety, and to handle corner cases in the future.

thescouser89 (approved):

PR #1243, opened 2024-09-18
[NCLSUP-1127] Implement switch to revert source naming to default fro…

jomrazek:
…m BC names

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
@@ -253,6 +253,12 @@ public class Pig {
                 description = "Path to file with license names for license generation. Usually rh-license-names.json.")
         private String licenseNamesPath;
 
+        @Option(
+                names = "--oldBCNaming",
+                defaultValue = "false",
+                description = "if set to true names used in source generation will be set in old way using BC names")
+        private boolean oldBCNaming;
+
         /**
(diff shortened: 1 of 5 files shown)


jomrazek:
@AntStephenson feel free to check and comment PR, can't add you to reviewer for some reason

AntStephenson:
LGTM, maybe because I'm not a maintainer

pig/src/main/java/org/jboss/pnc/bacon/pig/Pig.java
```diff
@@ -253,6 +253,12 @@ public static class Run extends PigCommand<PigRunOutput> {
                 description = "Path to file with license names for license generation. Usually rh-license-names.json.")
         private String licenseNamesPath;
 
+        @Option(
+                names = "--oldBCNaming",
+                defaultValue = "false",
+                description = "if set to true names used in source generation will be set in old way using BC names")
+        private boolean oldBCNaming;
```
janinko, line 260: Do we want to have the option here in general pig?

jomrazek: People usually run `pig run` to do all steps in pipeline, that's why it's here. Now if you're talking generally if we want switch and not just option for people to put it in build-config.yaml I would say yes because people will complain that they need to do code change in their files.

janinko (approved):

PR #1256, opened 2024-09-26
Exclude unwanted files from sources

jacobdotcosta:
Fixes https://issues.redhat.com/browse/QUARKUS-4940

### Checklist:


jomrazek:
This is fixed in NCLSUP-951 and discussed in NCLSUP-1126. Proper fix is to rebuild the affected artifacts and builds. I don't have problem with excluding this specific file but given we should not exclude anything from sources we should treat this as exception and not make it broader in the future for some reason.

AntStephenson:
I agree we should avoid excluding things from sources, but we should probably not ship this file so I think it's good to make an exception for this file only. Are you fine with this approving & merging this PR?

jomrazek (approved):

PR #1274, opened 2024-10-23
parse both dependency:tree and maven-dependency-plugin:tree

cunningt:
maven-dependency:tree has started showing up in the logs as dependency:tree rather than the elongated form of maven-dependency-plugin:tree - need to parse for both situations


jomrazek:
Hey, could you build your pr to reformat it, thanks.

cunningt:
> Hey, could you build your pr to reformat it, thanks.

Done!

PR #1327, opened 2025-03-03
NCL-6053: Prevent .bacon directories being created during tests. Migr…

rnc:
…ate testing library

Closes #459.

As #459 / NCL-6053 was never merged this provides an updated fix. This also migrates the test support library from system-lambda to https://github.com/webcompere/system-stubs


jomrazek (approved):
Looks good, thank you!

PR #1366, opened 2025-04-24
Add support for an offline license mirror

thescouser89:
The offline license mirror is found in licenses-generator/src/main/resources/offline-license-mirror and has the path structure of: `host/sub-path/sub-path2/text`

This is created to workaround issues we've encountered with rate-limits when trying to access www.gnu.org, blocking our license generation.

Since licenses don't change, it makes sense to cache them to:

1. workaround the rate-limit issue
2. speed up the license generation by avoiding downloading from the internet commonly used opensource license
3. 
Credit to Thomas Cunningham (@cunningt) who first mentioned the idea to me!

### Checklist:


thescouser89:
@cunningt @bekionel Please give it a review!

licenses-generator/src/main/java/org/jboss/pnc/bacon/licenses/xml/OfflineLicenseMirror.java
```diff
@@ -0,0 +1,39 @@
...
+
+        Path resourcesPathOfLicense = Paths.get(OFFLINE_LICENSE_MIRROR_RESOURCE_FOLDER, uri.getAuthority());
+
+        // consider each path in the url as a sub-folder
+        for (String subPath : uri.getPath().split("/")) {
+            resourcesPathOfLicense = resourcesPathOfLicense.resolve(subPath);
+        }
+
+        InputStream input = OfflineLicenseMirror.class.getClassLoader()
+                .getResourceAsStream(resourcesPathOfLicense.toString());
```
aloubyansky, line 36: While `Path.toString()` is OS-specific, classpath resource names must always use `/` as a separator.
So `uri.getPath()` could simply be re-used as-is to complete the classpath resource name.

thescouser89: Thanks for pointing it out! I had no idea. Resolved in commit: ceb9325bf4c73e647

aloubyansky:
Looks good to me @thescouser89, thanks! Would you like to squash the commits?

thescouser89:
Thanks @cunningt for the review! @bekionel let me know how the testing goes on your side!

PR #1431, opened 2025-07-17
[NCL-9108] Add backend option to cache artifacts downloaded from Indy

thescouser89:
Note: this feature is currently disabled in this commit until we can customize it on the frontend CLI side.

Generating huge maven repositories involve downloading a lot of artifacts from Indy. This is true for Maven repository generation types of:

- GENERATE
- BUILD_GROUP
- DOWNLOAD
- BUILD_CONFIGS
- MILESTONE

If we need to run the maven repository generations several times in the same machine, the same downloads happen and this can consume a lot of time. It would be nice if we could cache those downloads to speed up the maven repository generations for re-runs.

This commit adds the ability to cache the downloads into a cache file, managed my MapDB. The cache file is limited to a max disk size of 5GB, and the older contents in the cache will automatically get evicted.

The key of the cache is the download url, and the value is the byte array of the content of the downloaded file. The cache file location is in: `${user.home}/.cache/pnc-bacon/pnc-bacon-artifact-cache.db`.

It is currently disabled in the backend, but we'll expose a flag on the frontend in a future commit to enable/disable the cache feature.

### Checklist:

### DIFF
--- a/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileCache.java
+++ b/pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileCache.java
@@ -1,0 +1,72 @@
+package org.jboss.pnc.bacon.pig.impl.utils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.ConcurrentMap;
+
+import org.mapdb.DB;
+import org.mapdb.DBMaker;
+import org.mapdb.Serializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FileCache {
+
+    private static final Logger log = LoggerFactory.getLogger(FileCache.class);
+    private static final int MAX_SIZE_GB = 5;
+
+    private DB db;
+    private ConcurrentMap<String, byte[]> cachedMap;
+
+    public FileCache(Path cacheFile) {
+
+        // if cache file parent folder doesn't exist, create it
+        if (!Files.exists(cacheFile.getParent())) {
+            try {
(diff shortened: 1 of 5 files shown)


cunningt:
This sounds great!    Double checked how big our MRRC are and it looks like they are around 1GB -> 1.5GB.      I guess this doesn't apply to the RESOLVE_* generations though?     Is it possible to make those resolve from the cache?

cunningt (approved):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/FileCache.java
```diff
@@ -0,0 +1,71 @@
...
+    private ConcurrentMap<String, byte[]> cachedMap;
+
+    public FileCache(Path cacheFile) {
+
+        // if cache file parent folder doesn't exist, create it
+        if (!Files.exists(cacheFile.getParent())) {
+            try {
+                Files.createDirectories(cacheFile.getParent());
+            } catch (IOException e) {
+                log.error("Unable to create directory {}", cacheFile.getParent(), e);
```
mafh314, line 30: Hey @thescouser89 If the parent folder doesn't exist and the creation failed, shouldn't then return at this point?

thescouser89: Oh yeah you're right! Completely missed this. It should rethrow the error.

thescouser89:
@cunningt No for RESOLVE_ONLY. But the original description in NCL-9108 should in theory also help with that by adding a local m2 repository as a cache.

We first generate the maven repository from the BOM. We then copy that maven repository to the local m2 repository. Then on future generations, the local m2 repository will be checked before we download from remote repositories.

PR #1444, opened 2025-08-05
Add missing DA commands

jomrazek:
### Checklist:

### DIFF
--- a/da/src/main/java/org/jboss/bacon/da/DALookupCli.java
+++ b/da/src/main/java/org/jboss/bacon/da/DALookupCli.java
@@ -31,8 +31,14 @@ import org.jboss.da.lookup.model.MavenLatestRequest;
 import org.jboss.da.lookup.model.MavenLatestResult;
 import org.jboss.da.lookup.model.MavenLookupRequest;
 import org.jboss.da.lookup.model.MavenLookupResult;
+import org.jboss.da.lookup.model.MavenVersionsRequest;
+import org.jboss.da.lookup.model.MavenVersionsResult;
 import org.jboss.da.lookup.model.NPMLookupRequest;
 import org.jboss.da.lookup.model.NPMLookupResult;
+import org.jboss.da.lookup.model.NPMVersionsRequest;
+import org.jboss.da.lookup.model.NPMVersionsResult;
+import org.jboss.da.lookup.model.VersionDistanceRule;
+import org.jboss.da.lookup.model.VersionFilter;
 import org.jboss.da.model.rest.GAV;
 import org.jboss.da.model.rest.NPMPackage;
 import org.jboss.pnc.bacon.common.ObjectHelper;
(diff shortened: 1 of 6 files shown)


da/src/main/java/org/jboss/bacon/da/DALookupCli.java
```diff
@@ -111,6 +108,82 @@ public Integer call() {
...
+            description = "Lookup and filter available versions for the given Maven artifact coordinates (GAV).")
+    public static class LookupMavenVersions extends JSONCommandHandler implements Callable<Integer> {
+
+        @CommandLine.Option(
+                names = "--filter",
+                description = "Specifies what part of a version string should be used when filtering versions by " +
+                        "similarity. Version parts are: MAJOR.MINOR.MICRO.QUALIFIER-SUFFIX. Available filters: ALL,MAJOR,"
+                        +
+                        "MAJOR_MINOR,MAJOR_MINOR_MICRO,MAJOR_MINOR_MICRO_QUALIFIER")
+        private String filter;
```
akridl, line 122: **[subjective suggestion]** we can use here:
```java
private VersionFilter filter;
```

and can get rid of:
```java
.filter(VersionFilter.valueOf(filter))
```
but use:
```java
.filter(filter)
```
right away.

da/src/main/java/org/jboss/bacon/da/rest/endpoint/ListingsApi.java
```diff
@@ -53,58 +50,80 @@ public interface ListingsApi {
...
+            @QueryParam("artifactid") String artifactid,
+            @QueryParam("version") String version);
+
+    @DELETE
+    @Path("/blacklist/gav")
+    @Consumes({ "application/json" })
+    @Produces({ "application/json" })
+    SuccessResponse removeBlackArtifact(@Valid RestArtifact artifact);
+
+    @Deprecated(forRemoval = true)
```
akridl, line 79: don't we wanna remove these `/whitelist/*` endpoints completely? (since they are not present at DA server anymore, so an invocation to such endpoint leads to 500)

da/src/main/java/org/jboss/bacon/da/rest/endpoint/LookupApi.java
```diff
@@ -24,33 +24,49 @@
...
 
-import org.jboss.da.lookup.model.MavenLatestRequest;
-import org.jboss.da.lookup.model.MavenLatestResult;
-import org.jboss.da.lookup.model.MavenLookupRequest;
-import org.jboss.da.lookup.model.MavenLookupResult;
-import org.jboss.da.lookup.model.MavenVersionsRequest;
-import org.jboss.da.lookup.model.MavenVersionsResult;
-import org.jboss.da.lookup.model.NPMLookupRequest;
-import org.jboss.da.lookup.model.NPMLookupResult;
+import org.jboss.da.lookup.model.*;
```
akridl, line 27: :raised_eyebrow:

jomrazek: idea...

akridl (approved):
Thanks for adding the commands!!

PR #1455, opened 2025-08-19
Use force to regenerate autocompletion file

jomrazek:
This will regenerate bacon_completion file every build, otherwise it just complained file is there and won't generate it. If there is no change in commands, file is identical and not flagged in git changes.


akridl (approved):
LGTM!

PR #1461, opened 2025-08-28
[NCL-9254] Remove Lombok from Bacon

ppalaga:
### Checklist:

* Didn't add a note in the [CHANGELOG wiki](https://github.com/project-ncl/bacon/wiki/Changelog) because the change is not user-facing
* Haven't added unit tests, because it is a refactoring

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -19,10 +19,6 @@
             <groupId>org.jboss.pnc.bacon</groupId>
             <artifactId>config</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.projectlombok</groupId>
-            <artifactId>lombok</artifactId>
-        </dependency>
 
         <dependency>
             <groupId>org.keycloak</groupId>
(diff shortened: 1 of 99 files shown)


ppalaga:
The most of the changes were done using the following bash script:

```sh
#!/bin/bash

for f in $(grep -rl --include='*.java' 'import lombok.' .); do
  echo $f
  java -jar ~/.m2/repository/org/projectlombok/lombok/1.18.38/lombok-1.18.38.jar delombok -p "$f" \
    | sed -e '/Generated by delombok/d' -e '/lombok.Generated/d' \
    > "$f.new"
  mv "$f.new" "$f"
done
```

jomrazek:
Hi, can you provide some examples where lombok makes it hard to understand code from a first glance? I'm not telling I never debugged a code where I forgot some lombok annotation and spent significant time figuring out why it is not working but I would say that's not this case and pros of lombok are bigger, at least for me, than cons. I'm not in favor of this change especially in form of `@java.lang.Override    @java.lang.SuppressWarnings("all")` annotations all over the code. Refactored and with support of other developers, sure lets start the discussion.

PR #1487, opened 2025-09-25
NCL-9218 Add reqourConfig and simplify imports

rnc:
@jomrazek The idea here is to allow the bacon code to be more easily reusable without conflicting with Quarkus. By changing the imports to only whats needed it avoids bringing the rest-client module which which brings in resteasy. I have also added a section to be able to configure Reqour in future.


jomrazek (approved):
looks good, thank you

PR #1746, opened 2026-05-13
Remove whitelist from SCRG

jomrazek:
This removes whitelist functionality, it wasn't working for quite some time wrt data relevancy and completion. It will be replaced with labels and product info from PNC


akridl (approved):
LGTM, THX for removing this long non-used whitelisting feature!!

PR #1761, opened 2026-05-26
Update jboss parent to 53

jomrazek:
### Checklist:


pnc/pom.xml
```diff
@@ -10,6 +10,10 @@
     <artifactId>pnc</artifactId>
     <name>pnc</name>
 
+    <properties>
+        <jdk.min.version>17</jdk.min.version>
```
akridl: why is this needed? (we've already defined that in parent pom, didn't we?)

akridl (approved):

pnc/pom.xml
jomrazek: oh yeah

jomrazek: fixed, thanks!

PR #1792, opened 2026-07-15
Add jbang-catalog.json and document JBang usage in README

maxandersen:
Adds a JBang catalog so users can run/install bacon directly from Maven Central:

```shell
jbang bacon@project-ncl pnc build list
jbang app install bacon@project-ncl
```

**Changes:**
- `jbang-catalog.json` with aliases: `bacon`, `pnc`, `da`, `pig` (+ snapshot variants)
- README section documenting run, install,  and snapshot usage

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -1,0 +1,54 @@
+{
+  "aliases": {
+    "bacon": {
+      "script-ref": "org.jboss.pnc.bacon:cli:RELEASE:shaded",
+      "description": "PNC/DA/PiG CLI for ProjectNCL 2.0",
+      "java": "17+"
+    },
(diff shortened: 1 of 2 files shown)


jbang-catalog.json
```diff
@@ -0,0 +1,54 @@
+{
+  "aliases": {
+    "bacon": {
+      "script-ref": "org.jboss.pnc.bacon:cli:RELEASE:shaded",
+      "description": "PNC/DA/PiG CLI for ProjectNCL 2.0",
```
rnc, line 5: Why is this 2.0? Current version of Bacon is 3.5.0 and PNC is at 3.5.x. Would it not be better to remove the version?

PR #1800, opened 2026-07-23
Add skipDependencies option to build config generator

cunningt:
## Summary
- Adds a new `skipDependencies` boolean option to `BuildConfigGeneratorConfig` (defaults to `false`)
- When enabled, dependencies are cleared from the generated `PigConfiguration` output
- Documents the new option in the experimental guide

## Test plan
- [x] Verify build config generation works normally with `skipDependencies: false` (default)
- [x] Verify dependencies are omitted when `skipDependencies: true` is set
- [x] Verify existing behavior is unchanged when the option is not specified

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vibe13 (approved):
LGTM

PR #1802, opened 2026-07-29
Add automatic build environment and build tool detection

cunningt:
## Summary

- Adds automatic detection of JDK version, build tool (Maven vs Gradle), and build tool version for new build configs, replacing the previous hardcoded default environment
- Introduces a multi-source detection pipeline that inspects published JAR manifests on Maven Central, pom.xml/build.gradle from SCM, wrapper properties, .java-version files, and CI configs (Dockerfile, GitHub Actions, Jenkinsfile)
- Adds intelligent PNC environment selection that matches detected build info against available environments, ranking by exact version match > newer version > older version, with closest-version and simplest-environment tiebreaks

### Detection priority chain

1. **MANIFEST.MF** from Maven Central (`Build-Jdk-Spec` > `Build-Jdk` > `Created-By`)
2. **pom.xml** from SCM — `maven.compiler.release/source/target` properties, then `maven-compiler-plugin` configuration
3. **build.gradle** from SCM — toolchain (`JavaLanguageVersion.of`), `JavaVersion.VERSION_*`, `sourceCompatibility`
4. **Wrapper properties** — Maven/Gradle wrapper versions for build tool version detection
5. **`.java-version`** / **`.jenv/version`** files
6. **CI files** — Dockerfile, GitHub Actions workflows, Jenkinsfile

Falls back to JDK 11 + Maven if nothing is detected.

### New classes

- `ProjectBuildInfoDetector` — core detection engine
- `ProjectBuildInfo` — data class for detected build metadata
- `JdkVersion` — enum for supported JDK versions (1.8, 11, 17, 21, 25) with version string parsing
- `ScmFileAccessor` — fetches individual files from GitHub/GitLab repos via raw-content APIs
- `PncEnvironmentGenerator` — standalone utility to fetch PNC environments to JSON
- `EnvironmentResolver.selectEnvironment()` — new method for intelligent environment matching

## Test plan

- [x] `JdkVersionTest` — parameterized tests for version string parsing and round-down behavior
- [x] `ProjectBuildInfoDetectorTest` — manifest parsing, pom.xml parsing, Gradle build file parsing
- [x] `EnvironmentResolverTest` — JDK matching, build type filtering, tool version ranking, deprecation/hidden exclusion, fallback behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1806, opened 2026-08-03
Add automatic build environment and build tool detection - #1802 +  additional enforcers checks

vibe13:
## Summary

This PR builds on project-ncl/bacon#1802 and refines automatic PNC build
environment selection.

It:

- distinguishes preferred wrapper versions from hard Maven requirements;
- combines Maven requirements without losing upper bounds;
- selects the least disruptive compatible environment;
- avoids Maven or Gradle downgrades when replacing an environment;
- prefers simple environments over bundled toolchain environments;
- retains an active existing environment when it is already compatible;
- optionally reselects environments for existing BuildConfigs;
- keeps a conclusively detected project JDK;
- uses the existing BuildConfig JDK only when project detection fell back
  to the configured default.

## Commit structure

1. Detect Maven preferences and hard requirements.
2. Select the least disruptive compatible environment.
3. Optionally reselect existing BuildConfig environments.


jomrazek (approved):
👌

PR #1807, opened 2026-08-03
NCLSUP-1459 Add retry for DA server calls

cunningt:
https://redhat.atlassian.net/browse/NCLSUP-1459

## Summary
- Adds `executeWithRetry()` to `DaHelper` with exponential backoff for transient failures (`ProcessingException`, 5xx responses), immediate failure on SSL certificate errors and 4xx responses, and `FatalException` wrapping after retries are exhausted
- Wraps all DA lookup calls in `DALookupCli` with the new retry mechanism
- Adds explicit connect (30s) and read (60s) timeouts on the DA REST client
- Adds configurable `daMaxRetries` to `DependencyResolutionConfig` and propagates it through `DependencyResolver` and `ProjectFinder` in the experimental module
- Includes comprehensive unit tests for retry behavior (success, transient failures, SSL errors, 4xx/5xx, max retries)

## Test plan
- [x] Verify `DaHelperTest` passes all new retry test cases
- [x] Verify `ProjectFinderTest` passes with updated constructor
- [x] Test DA lookups against a live DA server to confirm retry behavior on transient failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/da/src/main/java/org/jboss/bacon/da/DaHelper.java
+++ b/da/src/main/java/org/jboss/bacon/da/DaHelper.java
@@ -21,7 +21,13 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
+
+import javax.net.ssl.SSLHandshakeException;
+import javax.ws.rs.ProcessingException;
+import javax.ws.rs.WebApplicationException;
 
 import org.jboss.bacon.da.rest.endpoint.ListingsApi;
 import org.jboss.bacon.da.rest.endpoint.LookupApi;
@@ -35,6 +41,7 @@ import org.jboss.pnc.bacon.auth.client.PncClientHelper;
 import org.jboss.pnc.bacon.common.CustomRestHeaderFilter;
 import org.jboss.pnc.bacon.common.TokenAuthenticator;
 import org.jboss.pnc.bacon.common.Utils;
+import org.jboss.pnc.bacon.common.exception.FatalException;
 import org.jboss.pnc.bacon.config.Config;
 import org.jboss.pnc.bacon.config.DaConfig;
 import org.jboss.pnc.client.Configuration;
@@ -46,11 +53,17 @@ import org.jboss.resteasy.plugins.providers.RegisterBuiltin;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 import io.opentelemetry.api.trace.Span;
+import lombok.extern.slf4j.Slf4j;
 
 /**
  * Helper methods for DA stuff
  */
+@Slf4j
 public class DaHelper {
+
+    private static final int DEFAULT_MAX_RETRIES = 5;
+    private static final long CONNECT_TIMEOUT_SECONDS = 30;
+    private static final long READ_TIMEOUT_SECONDS = 60;
     private final static String DA_PATH = "/rest/v-1";
 
     private static ResteasyClientBuilder builder;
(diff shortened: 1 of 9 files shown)


da/src/main/java/org/jboss/bacon/da/DaHelper.java
```diff
@@ -230,4 +245,74 @@ public static <T extends NPMResult> List<T> orderNPMResult(
...
+                if (retries > maxRetries) {
+                    throw new FatalException(
+                            "DA call failed after " + maxRetries + " retries: " + operationDescription,
+                            e);
+                }
+                logRetryAttempt(retries, maxRetries, operationDescription, e);
+                sleepExponentially(retries);
+            } catch (WebApplicationException e) {
+                int statusCode = e.getResponse().getStatus();
+                if (statusCode >= 500) {
```
vibe13: I would filter the retries (with exponential backoff)  to some specific statuses:
```
408 Request Timeout
429 Too Many Requests
500 Internal Server Error
502 Bad Gateway
503 Service Unavailable
504 Gateway Timeout
```
The first 2 are 4xx but should be safe to retry.

All 5xx might be not safe / sensible to retry:
```
501 Not Implemented
505 HTTP Version Not Supported
```
Same for:
```
506 Variant Also Negotiates
507 Insufficient Storage
508 Loop Detected
510 Not Extended
511 Network Authentication Required
```
should probably be avoided because a short delay might unlikely fix those .

jomrazek (approved):
Looks good

da/src/main/java/org/jboss/bacon/da/DaHelper.java
vibe13: Once the status code is identified and either wrapped in the `FatalException` or retried, the response is discarded, so I would close the response straight away:

```
if (statusCode >= 500) {
        e.getResponse().close();
```

da/src/main/java/org/jboss/bacon/da/DaHelper.java
```diff
@@ -230,4 +245,74 @@ public static <T extends NPMResult> List<T> orderNPMResult(
...
+
+    private static void logRetryAttempt(int retries, int maxRetries, String operationDescription, Exception e) {
+        if (retries == maxRetries / 2) {
+            log.warn("Having difficulty reaching DA server for {}. Retrying...", operationDescription);
+        }
+        log.debug("DA Retry {}/{} for {}: {}", retries, maxRetries, operationDescription, e.getMessage());
+    }
+
+    private static void sleepExponentially(int retries) {
+        long amountOfSleep = (long) (100 * Math.pow(2, retries));
```
vibe13: I would cap the max backoff so that we avoid situation where we increased the retries but overlooked the exponential effect, and add some jitter:

```
long exponentialDelay = calculateExponentialBackoffMillis(retries);
long jitteredDelay = jitter.applyAsLong(exponentialDelay);
amountOfSleep = Math.max(0, Math.min(jitteredDelay, MAX_BACKOFF_MILLIS));
```

```
private static final long INITIAL_BACKOFF_MILLIS = 200;
private static final long MAX_BACKOFF_MILLIS = 5_000;

private static long calculateExponentialBackoffMillis(int retries) {
    long delay = INITIAL_BACKOFF_MILLIS;

    for (int retry = 1; retry < retries; retry++) {
        if (delay >= MAX_BACKOFF_MILLIS / 2) {
            return MAX_BACKOFF_MILLIS;
        }

        delay *= 2;
    }

    return Math.min(delay, MAX_BACKOFF_MILLIS);
}
private static long applyJitter(long delayMillis) {
    long lowerBound = delayMillis / 2;
    long upperBoundExclusive = delayMillis + 1;

    return ThreadLocalRandom.current()
            .nextLong(lowerBound, upperBoundExclusive);
}
```

vibe13 (changes requested):
I Tom, this is a big improvement! I suggested some changes bases on previous experience in other services. 
408 and 429 responses should also provide a `Retry-After` which could be used, but this is definitely good enough as a big improvement on the reliability.
Thanks for working on this!

da/src/main/java/org/jboss/bacon/da/DaHelper.java
cunningt: Thank you for the suggestion - added!

da/src/main/java/org/jboss/bacon/da/DaHelper.java
cunningt: Added!

cunningt: This is a good feedback - the most common situation I've been encountering here is a lack of response (no status code)

vibe13 (approved):

PR #1822, opened 2026-08-25
Implement PNC retries

xstefank:
-


jomrazek:
Hi Martin, thank you for your contribution, it's true we accidentally removed the retries for pnc client when changing the client provider. We found out and implemented it back few days ago https://github.com/project-ncl/pnc/commit/f953e513c7b0053cbb986ecccdae7d5412f0adbc and we're already using this change in bacon snapshot. I believe you have some usecase where timeouts happened when you created this change. Can you use snapshot version and check if our client change resolves issues you're seeing or if your change is also necessary? Thank you

PR #1825, opened 2026-08-31
Change to artifactory repo provider

jomrazek:
### Checklist:


pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/repository/RepositoryProvider.java
```diff
@@ -0,0 +1,64 @@
...
+public class RepositoryProvider {
+    private static volatile String repoProvUrl;
+    private static volatile String tempRepoProvUrl;
+
+    private RepositoryProvider() {
+    }
+
+    public static String getRepoProviderUrl() {
+        if (repoProvUrl == null) {
+            repoProvUrl = pigUrl() + "artifactory/pnc-mvn-builds-imports-public";
```
jomrazek: @michalovjan @pkocandr can you please check if these groups correctly represent old indy goups, for this instance it was indy "api/content/maven/group/static" group

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/repository/RepositoryProvider.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        return repoProvUrl;
+    }
+
+    public static String getTempRepoProviderUrl() {
+        if (tempRepoProvUrl == null) {
+            tempRepoProvUrl = pigUrl() + "artifactory/pnc-mvn-temp-ibm-builds";
```
jomrazek: @michalovjan @pkocandr second case, for this instance it was indy "api/content/maven/hosted/temporary-builds" group. Thank you

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/repository/RepositoryProvider.java
pkocandr: An equivalent for `static` group is going to be `pnc-mvn-builds-imports` which does not exist yet.

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/repository/RepositoryProvider.java
pkocandr: This should be `pnc-mvn-ibm-temp-builds` (you have `pnc-mvn-temp-ibm-builds`).

pkocandr (changes requested):

pig/src/main/java/org/jboss/pnc/bacon/pig/impl/utils/repository/RepositoryProvider.java
michalovjan: This won't work across different environments. Some pig functionality will be broken if we use `stage` and `devel` PNC instances.

The repos on `stage` begin with `pnc-stage` project and on `devel` with `pnc-devel`.

jomrazek: @michalovjan yup this is fixed in up to date PR version

jomrazek:
Merging this for snapshot testing, group will be fixed when available

PR #1831, opened 2026-09-03
Update the regex for rorcksdbjni and use a different alias of CDDL in javax.annotation-api exclusion

vardhmanJain:


rgdoliveira:
LGTM. @thescouser89 if you can check that one please.

rgdoliveira:
the fail in CI does not seem related to the change in this PR

